PR #47, opened 2017-12-27
Serde support

### LINKED ISSUE #37: Add support for Serde
Add support controlled by a `serde` feature. Enable or disable by default?

https://crates.io/crates/serde

radix:
Implement Serialize and Deserialize for Quantity, with tests.

Fixes #37.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -54,8 +54,8 @@ install:
 ## Build Script ##
 build: false
 test_script:
-  - cargo build --verbose
-  - cargo test --verbose
+  - cargo build --verbose --all-features --tests
+  - cargo test --verbose --features serde
 
 notifications:
     - provider: Email
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 ### Added
  * [#26](https://github.com/iliekturtles/uom/issues/26) Implement `num::Zero`.
  * [#35](https://github.com/iliekturtles/uom/issues/35) Implement `num::Saturating`.
+ * [#37](https://github.com/iliekturtles/uom/issues/35) Implement `serde::Serialize` and
+   `serde::Deserialize`. Disabled by default. Enabled with the `use_serde` feature.
 
 ## [v0.16.0] — 2017-12-21
 This release contains significant changes in order to support underlying storage types that
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,15 +22,23 @@ maintenance = { status = "actively-developed" }
 
 [dependencies]
 num = "0.1"
+serde = { version = "1.0", optional = true, default-features = false }
 typenum = "1.9.0"
 
 [dev-dependencies]
 approx = "0.1.1"
 quickcheck = "0.5.0"
+serde_json = "1.0"
 static_assertions = "0.2.5"
 
 [features]
 default = ["f32", "f64", "si", "std"]
+# The `use_serde` feature exists so that, in the future, we can add other dependency features to
+# it, like num/serde. However, num/serde is currently left out because it has not yet been updated
+# to Serde 1.0.
+# It is also necessary to call it something other than `serde` because of this cargo bug:
+# https://github.com/rust-lang/cargo/issues/1286
+use_serde = ["serde"]
 usize = []
 u8 = []
 u16 = []
--- a/README.md
+++ b/README.md
@@ -63,10 +63,10 @@ See the [examples](examples) directory for more advanced usage:
 
 ## Features
 `uom` has multiple `Cargo` features for controlling available underlying storage types, the
-inclusion of the pre-built [International System of Units][si] (SI), and `no_std` functionality. The
-features are described below. `f32`, `f64`, `std`, and `si` are enabled by default. Features can be
-cherry-picked by using the `--no-default-features` and `--features "..."` flags when compiling `uom`
-or specifying features in Cargo.toml:
+inclusion of the pre-built [International System of Units][si] (SI), support for [Serde][serde],
+and `no_std` functionality. The features are described below. `f32`, `f64`, `std`, and `si` are
+enabled by default. Features can be cherry-picked by using the `--no-default-features` and
+`--features "..."` flags when compiling `uom` or specifying features in Cargo.toml:
 
 ```toml
 [dependencies]
@@ -74,6 +74,7 @@ uom = {
     version = "0.16.0",
     default-features = false,
     features = [
+        "use_serde", # Serde support.
         "usize", "u8", "u16", "u32", "u64", # Unsigned integer storage types.
         "isize", "i8", "i16", "i32", "i64", # Signed interger storage types.
         "bigint", "biguint", # Arbitrary width integer storage types.
@@ -88,12 +89,15 @@ uom = {
    `rational`, `rational32`, `rational64`, `bigrational`, `f32`, `f64` -- Features to enable
    underlying storage types. At least one of these features must be enabled. `f32` and `f64` are
    enabled by default.
+ * `use_serde` -- Feature to enable support for serialization and deserialization of quantities
+   with the [serde][serde] crate. Disabled by default.
  * `si` -- Feature to include the pre-built [International System of Units][si] (SI). Enabled by
    default.
  * `std` -- Feature to compile with standard library support. Disabling this feature compiles `uom`
    with `no_std`. Enabled by default.
 
 [si]: http://jcgm.bipm.org/vim/en/1.16.html
+[serde]: https://serde.rs/
 
 ## Design
 Rather than working with [measurement units](http://jcgm.bipm.org/vim/en/1.9.html) (meter,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,10 +52,11 @@
 //!
 //! ## Features
 //! `uom` has multiple `Cargo` features for controlling available underlying storage types, the
-//! inclusion of the pre-built [International System of Units][si] (SI), and `no_std` functionality.
-//! The features are described below. `f32`, `f64`, `std`, and `si` are enabled by default. Features
-//! can be cherry-picked by using the `--no-default-features` and `--features "..."` flags when
-//! compiling `uom` or specifying features in Cargo.toml:
+//! inclusion of the pre-built [International System of Units][si] (SI), support for
+//! [Serde][serde], and `no_std` functionality. The features are described below. `f32`, `f64`,
+//! `std`, and `si` are enabled by default. Features can be cherry-picked by using the
+//! `--no-default-features` and `--features "..."` flags when compiling `uom` or specifying
+//! features in Cargo.toml:
 //!
 //! ```toml
 //! [dependencies]
@@ -63,6 +64,7 @@
 //!     version = "0.16.0",
 //!     default-features = false,
 //!     features = [
+//!         "use_serde", # Serde support.
 //!         "usize", "u8", "u16", "u32", "u64", # Unsigned integer storage types.
 //!         "isize", "i8", "i16", "i32", "i64", # Signed interger storage types.
 //!         "bigint", "biguint", # Arbitrary width integer storage types.
@@ -77,12 +79,15 @@
 //!    `rational`, `rational32`, `rational64`, `bigrational`, `f32`, `f64` -- Features to enable
 //!    underlying storage types. At least one of these features must be enabled. `f32` and `f64` are
 //!    enabled by default.
+//!  * `use_serde` -- Feature to enable support for serialization and deserialization of quantities
+//!    with the [serde][serde] crate. Disabled by default.
 //!  * `si` -- Feature to include the pre-built [International System of Units][si] (SI). Enabled by
 //!    default.
 //!  * `std` -- Feature to compile with standard library support. Disabling this feature compiles
 //!    `uom` with `no_std`. Enabled by default.
 //!
 //! [si]: http://jcgm.bipm.org/vim/en/1.16.html
+//! [serde]: https://serde.rs/
 //!
 //! ## Design
 //! Rather than working with [measurement units](http://jcgm.bipm.org/vim/en/1.9.html) (meter,
@@ -157,6 +162,10 @@ compile_error!("A least one underlying storage type must be enabled. See the fea
 pub extern crate num;
 
 #[doc(hidden)]
+#[cfg(feature = "serde")]
+pub extern crate serde;
+
+#[doc(hidden)]
 pub extern crate typenum;
 
 #[cfg(all(test, any(feature = "f32", feature = "f64")))]
@@ -165,6 +174,8 @@ extern crate approx;
 #[cfg(test)]
 #[macro_use]
 extern crate quickcheck;
+#[cfg(all(test, feature = "serde"))]
+extern crate serde_json;
 #[cfg(test)]
 #[macro_use]
 extern crate static_assertions;
--- a/src/system.rs
+++ b/src/system.rs
@@ -873,6 +873,42 @@ macro_rules! system {
             }
         }
 
+        #[cfg(feature = "serde")]
+        impl<D, U, V> $crate::serde::Serialize for Quantity<D, U, V>
+        where
+            D: Dimension + ?Sized,
+            U: Units<V> + ?Sized,
+            V: $crate::num::Num + $crate::Conversion<V> + $crate::serde::Serialize,
+        {
+            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+            where
+                S: $crate::serde::Serializer
+            {
+                self.value.serialize(serializer)
+            }
+        }
+
+        #[cfg(feature = "serde")]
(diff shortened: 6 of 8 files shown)


Cargo.toml
```diff
@@ -23,6 +23,7 @@ maintenance = { status = "actively-developed" }
 [dependencies]
 num = "0.1"
 typenum = "1.9.0"
+serde = { version = "1.0", optional = true, features = ["derive"] }
```
iliekturtles: Move above `typenum` so dependencies are in alphabetical order.

src/lib.rs
```diff
@@ -155,6 +155,9 @@ compile_error!("A least one underlying storage type must be enabled. See the fea
 
 #[doc(hidden)]
 pub extern crate num;
```
iliekturtles: Add a blank line just below `num`.

src/system.rs
```diff
@@ -858,6 +858,38 @@ macro_rules! system {
...
         }
 
+        #[cfg(feature = "serde")]
+        impl<D, U, V> $crate::serde::ser::Serialize for Quantity<D, U, V>
+        where
+            D: Dimension + ?Sized,
+            U: Units<V> + ?Sized,
+            V: $crate::num::Num + $crate::Conversion<V> + $crate::serde::ser::Serialize,
+        {
+            fn serialize<S: $crate::serde::ser::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
```
iliekturtles: Put `S`'s type bounds in a `where` clause instead of inline.

src/system.rs
```diff
@@ -858,6 +858,38 @@ macro_rules! system {
...
+        where
+            D: Dimension + ?Sized,
+            U: Units<V> + ?Sized,
+            V: $crate::num::Num + $crate::Conversion<V> + $crate::serde::de::Deserialize<'de>,
+        {
+            fn deserialize<De>(deserializer: De) -> Result<Self, De::Error>
+            where
+                De: $crate::serde::de::Deserializer<'de>,
+            {
+                let value: V = $crate::serde::de::Deserialize::deserialize(deserializer)?;
```
iliekturtles: Blank line beneath this one.

src/lib.rs
```diff
@@ -155,6 +155,9 @@ compile_error!("A least one underlying storage type must be enabled. See the fea
...
+#[doc(hidden)]
```
iliekturtles: I think this should be `#[cfg_attr(feature = "serde", doc(hidden)]` so that the attribute only appears when `serde` is enabled. Didn't test with the compiler however.

radix: I tested running `cargo doc` with and without the `serde` feature enabled and it seemed to work fine in both cases (not showing docs for serde in either case). However, of course, if you'd like me to specify a cfg_attr anyway I can do that.

radix:
I've addressed your comments, but I still need to write some tests. I should be able to get to that tomorrow! :)

radix:
Okay, this is probably ready for another review pass. I've implemented simple tests. Note that I've also set up travis to run the serde tests in addition to the default-feature tests.

radix:
Oh, I'll also update CHANGELOG.md.

src/lib.rs
```diff
@@ -156,6 +159,13 @@ compile_error!("A least one underlying storage type must be enabled. See the fea
 #[doc(hidden)]
 pub extern crate num;
 
+#[doc(hidden)]
```
iliekturtles: Looks like `cfg_attr` isn't necessary. The compiler is smart enough that `#[doc(hidden)]` doesn't "fall through" to the next item when the `serde` feature is not enabled.

src/lib.rs
```diff
@@ -156,6 +159,13 @@ compile_error!("A least one underlying storage type must be enabled. See the fea
...
+#[cfg(feature = "serde")]
+pub extern crate serde;
+
+#[cfg(all(feature = "serde", test))]
```
iliekturtles: Change order so that `test` appears first. `#[cfg(all(test, feature = "serde"))]`. I also have the dev-dependencies in a separate section beneath the regular dependencies. Not sure if this should be maintained or if all dependencies should just be included in order. Thoughts?

src/tests.rs
```diff
@@ -7,6 +7,8 @@ use num::{Float, FromPrimitive, One, Saturating, Signed, Zero};
 use quickcheck::TestResult;
 use lib::fmt::Debug;
 use lib::marker::PhantomData;
+#[cfg(all(feature = "serde", test))]
```
iliekturtles: `test` isn't necessary since this module is only compiled when `test` is set.

src/tests.rs
```diff
@@ -527,6 +529,26 @@ mod system_macro {
...
+                let json_q = serde_json::to_string(&m).expect("Must be able to serialize Quantity");
+                let json_f = serde_json::to_string(&*v).expect("Must be able to serialize num");
+
+                json_f == json_q
+            }
+
+            #[cfg(feature = "serde")]
+            #[allow(trivial_casts)]
+            fn serde_deserialize(v: A<V>) -> bool {
+                let json_f = serde_json::to_string(&*v).expect("Must be able to deserialize num");
```
iliekturtles: I believe the `expect(...)` should say serialize and not deserialize.

src/tests.rs
```diff
@@ -527,6 +529,26 @@ mod system_macro {
                         &(Length::new::<meter>((*l).clone())
                             % Length::new::<meter>((*r).clone())).get(meter)))
             }
+
+            #[cfg(feature = "serde")]
+            #[allow(trivial_casts)]
+            fn serde_serialize(v: A<V>) -> bool {
+                let m = Length::new::<meter>((*v).clone());
+                let json_q = serde_json::to_string(&m).expect("Must be able to serialize Quantity");
```
iliekturtles: Put `let json_q = ...` after `let json_f = ...` to match comparison order below.

src/tests.rs
```diff
@@ -527,6 +529,26 @@ mod system_macro {
...
+            }
+
+            #[cfg(feature = "serde")]
+            #[allow(trivial_casts)]
+            fn serde_deserialize(v: A<V>) -> bool {
+                let json_f = serde_json::to_string(&*v).expect("Must be able to deserialize num");
+                let length: Length = serde_json::from_str(&json_f)
+                    .expect("Must be able to deserialize Quantity");
+
+                Test::approx_eq(&length.get(meter), &*v)
```
iliekturtles: Make `&*v` the first parameter.

README.md
```diff
@@ -63,10 +63,10 @@ See the [examples](examples) directory for more advanced usage:
...
 ## Features
 `uom` has multiple `Cargo` features for controlling available underlying storage types, the
-inclusion of the pre-built [International System of Units][si] (SI), and `no_std` functionality. The
-features are described below. `f32`, `f64`, `std`, and `si` are enabled by default. Features can be
-cherry-picked by using the `--no-default-features` and `--features "..."` flags when compiling `uom`
-or specifying features in Cargo.toml:
+inclusion of the pre-built [International System of Units][si] (SI), support for [Serde][serde], and
+`no_std` functionality. The features are described below. `f32`, `f64`, `std`, and `si` are enabled
+by default. Features can be cherry-picked by using the `--no-default-features` and
+`--features "..."` flags when compiling `uom` or specifying features in Cargo.toml:
```
iliekturtles: Wrap at 100 characters.

README.md
```diff
@@ -88,12 +88,15 @@ uom = {
    `rational`, `rational32`, `rational64`, `bigrational`, `f32`, `f64` -- Features to enable
    underlying storage types. At least one of these features must be enabled. `f32` and `f64` are
    enabled by default.
+ * `serde` -- Feature to enable support for serialization and deserialization of quantities with
+   the [serde][serde] crate.
```
iliekturtles: Note that the feature is disabled by default. Also include the `serde` feature in the `toml` block above: `"serde", # Serde support.`

src/lib.rs
radix: I'm happy to maintain the current style, I'll move it below.

src/tests.rs
radix: Right you are!

radix:
I believe I've addressed your comments, thanks for the review!

radix:
FWIW, the appveyor failures are only in the old `rem` tests that occasionally fail, not the new serde code.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
 ### Added
  * [#26](https://github.com/iliekturtles/uom/issues/26) Implement `num::Zero`.
  * [#35](https://github.com/iliekturtles/uom/issues/35) Implement `num::Saturating`.
+ * [#37](https://github.com/iliekturtles/uom/issues/35) Implement `serde::Serialize` and
+   `serde::Deserialize`. Disabled by default. Enabled with the `serde` cargo feature.
```
iliekturtles: Just say "`serde` feature".

README.md
```diff
@@ -79,6 +79,7 @@ uom = {
         "bigint", "biguint", # Arbitrary width integer storage types.
         "rational", "rational32", "rational64", "bigrational", # Integer ratio storage types.
         "f32", "f64", # Floating point storage types.
+	"serde", # Serde support.
```
iliekturtles: Looks like this line didn't get indented.

src/tests.rs
```diff
@@ -7,6 +7,8 @@ use num::{Float, FromPrimitive, One, Saturating, Signed, Zero};
 use quickcheck::TestResult;
 use lib::fmt::Debug;
 use lib::marker::PhantomData;
+#[cfg(all(feature = "serde"))]
```
iliekturtles: `all(...)` isn't needed anymore.

.appveyor.yml
```diff
@@ -56,6 +56,7 @@ build: false
 test_script:
   - cargo build --verbose
   - cargo test --verbose
```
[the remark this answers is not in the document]
radix: That's fair. I should say that I did it this way intentionally, to make sure that the crate still builds and tests correctly *without* the serde feature enabled (making sure I set up my `cfg` attributes correctly etc). But with enough of these kinds of features it would lead to ridiculous test times, having to run through the test suite several times... so yeah, I can remove it :)

README.md
radix: oops, that was a sneaky tab character.

iliekturtles:
I spoke a bit too soon about being ready to merge. I finally read through the Serde documentation and have a few more questions:

- Should we be using `serialize_struct` or `serialize_newtype_struct`: https://serde.rs/impl-serialize.html#serializing-a-struct
- Should we be using this instead of `serde_json`: https://serde.rs/unit-testing.html
- https://serde.rs/feature-flags.html#-feature-std -- I believe we should always disable this feature since we don't need the `impl`s it provides. Or possibly tie to `uom`'s `std` feature.

radix:
@iliekturtles no problem!

1. Hmm, that's interesting. so, to get the first thing out of the way -- not `serialize_struct`, because we're not serializing a compound structure here, only one value (the bare number in the base unit).

As for `serialize_newtype_struct` -- I believe what I implemented is equivalent to using `serialize_newtype_struct`, and honestly I'm not sure what the point of `serialize_newtype_struct` is. As the documentation says, implementors of serialization formats are supposed to just serialize the wrapped value, but I suppose it gives them the option of doing something different (against the advice of the documentation). I'm honestly not sure why I should use it -- I'm currently trying to get some advice from someone who is more familiar with it. It's worth noting that our struct isn't literally a newtype, since it has those other phantom fields.  Looking at the implementations of backends, it seems like `serializer.serialize_newtype_struct("Foobar", &self.0)` is always equivalent to `self.0.serialize(serializer)`.

2. I've always just used serde_json personally, since it's simpler and more concise to use than serde_test. I think serde_test might be more useful when you want somewhat more specific tests (e.g. testing the order of serialization of fields and whatnot). However, I'd be happy to switch -- it should be simple enough.

3. Good point. I'll see if there are any complications...

radix:
Okay, about point 1: https://github.com/ron-rs/ron serializes "serializes_newtype_struct" differently, so the name you pass is actually encoded into the output. I could also perhaps pass the name of the base unit as the newtype name, to make that output a bit more descriptive. I'll also have to deserialize with deserialize_newtype_struct. I'll make that change and let you know!

iliekturtles:
Thanks for the information. Leave points 1 and 2 as-is for the moment. I'm going to do a bit more digging.

radix:
@iliekturtles Just FYI I was looking into using serialize_newtype_struct -- that's easy enough, but the deserialization side is more complicated. It needs a visitor to be passed, and I was trying to figure out how to generate one of those without too much work. But okay, I'll leave that alone and get back to disabling serde's `std` feature -- which has its own complications, because apparently our dependency on other crates is relying on the serde std feature. I'm digging into that some more.

radix:
Okay, I think I fixed the third point, but I was real confused for a while. Basically, there is a cargo bug when building uom directly with serde enabled: https://github.com/rust-lang/cargo/issues/4664

Basically, because we have a dev-dependency on serde_json, which depends on serde (with default features), cargo commands run *directly against uom* still build serde with default features. However, when uom is used as a dependency, uom's dev-dependencies is ignored and so serde won't require std. I verified this manually by creating an example project that depends on my fork of uom, with serde enabled. I built it with --verbose and saw that it does not include serde's std feature.

radix:
btw, @iliekturtles , if you want to chat about any of these things in real-time, I am on the various rust-related channels on the mozilla IRC channels as `M-radix` (or `radix`)

iliekturtles:
```
error[E0277]: the trait bound `num::BigInt: serde::Serialize` is not satisfied
   --> src\tests.rs:537:30
    |
537 |                 let json_f = serde_json::to_string(&*v).expect("Must be able to serialize num");
    |                              ^^^^^^^^^^^^^^^^^^^^^ the trait `serde::Serialize` is not implemented for `num::BigInt`
    |
    = note: required by `serde_json::to_string`
```

One more issue I found when running `cargo test --all-features`. `BigInt` and other complex types don't implement `serde` traits by default. Looks like there are features to include this functionality: https://github.com/rust-num/num/blob/master/Cargo.toml. Note that I just pushed a fix for `--all-features` not compiling because of `num::Saturating`.

Once this is resolved I'll merge for real. I still want to look at points 1 and 2 above, but will complete that before releasing v0.17.0.

I used to be on IRC via a very old laptop that was lost to a storm this fall. I'll see about getting connected again.

radix:
Ooookay. Unfortunately I think this is going to cause a bit of a pain, because of this bug: https://github.com/rust-lang/cargo/issues/1286

I think what we want is to define a feature that both enables a dependency (serde) *and* enables features in other dependencies (num-bigint/serde, et al). But because of that linked bug, we cannot name this feature "serde", because all optional dependencies have names equal to the package name -- and sadly cargo doesn't just merge the features in the `features` section with the ones from the dependencies section.

So this means we need to bikeshed a name: what should it be? `use_serde`? `enable_serde`?

radix:
Ah! Even more confounding problems! :angry:

Apparently the `num` crate has not been updated to serde 1.0. So there's no hope for us to provide serde support for its data types yet. https://github.com/rust-num/num/issues/357

So I suppose we should just disable the tests for those backend types for now. However, I suppose we should still standardize/document a feature like `use_serde` instead of `serde`, so that we can later add `num/serde` to that feature once they update their support to serde 1.0.


So I'll

1. add the `use_serde` feature and document it as the way to turn on serde support
2. move the tests for serde into a new submodule that has everything enabled *except* the num datatypes.
3. how about for now I also add a `cargo build --all-features --tests` to the travis/appveyor configs, so that we at least know the code compiles, if not passes tests?

iliekturtles:
1. `use_serde` sounds good. 
1. Move the `serde` tests into the `op_assign` module and rename to `mod primitive`.
1. Good idea. Change AppVeyor and TravisCI to `cargo build --all-features --verbose`. As tests start getting fixed we can add additional features until build servers are running `cargo test --all-features`.

radix:
Okay, sorry for the delay @iliekturtles ! I think I've addressed the outstanding comments.

Oh, one minor difference which I'm not even sure was intentional on your part: I changed the build command to `cargo build --verbose --all-features --tests` instead of just `cargo build --verbose --all-features`. This means that the test code is at least compiled (but not necessarily run) for all features -- that's what I was originally suggesting in https://github.com/iliekturtles/uom/pull/47#issuecomment-354511514

iliekturtles:
Merged, thanks! I ended up fixing a formatting issue in `tests.rs` and re-organizing some comments. Also changed build script back to `cargo build --all-features`. Including `--tests` causes Cargo to do nothing for some reason (https://travis-ci.org/iliekturtles/uom/jobs/324219347#L578).

radix:
Hmm, well that's definitely strange. As you can see, other builds *do* show output, even from the same travis run as you linked. I wonder if there is just some caching going on that's causing cargo to realize it has nothing to do.

iliekturtles:
I get the same thing locally which is where I found the issue originally:

```
uom (master)$ cargo clean && cargo build --all-features --tests
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
```

PR #65, opened 2018-04-11
Add separate testcrate

Nemo157:
Allows testing without inheriting features from the uom build, is tested
when running `cargo test` automatically be being a workspace member, so
will be tested on travis and appveyor. Includes all the examples from
uom so that these examples are tested as if they were part of a separate
application.

See previous discussion at #62. This is used for testing whether #61 is fixed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,10 @@ travis-ci = { repository = "iliekturtles/uom" }
 coveralls = { repository = "iliekturtles/uom" }
 maintenance = { status = "actively-developed" }
 
+[workspace]
+members = ["testcrate"]
+default-members = [".", "testcrate"]
+
 [dependencies]
 num = "0.1"
 serde = { version = "1.0", optional = true, default-features = false }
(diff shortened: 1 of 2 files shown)


iliekturtles:
Per the discussion in #62 I'd prefer to go the route of using `compiletest-rs` since it doesn't require switching over to a workspace. Although as I also mentioned in the other PR `uom` may eventually need to switch to a workspace if custom derive is used. Thoughts?

Nemo157:
Yeah `compiletest-rs` looks interesting and probably a better solution to this. I haven't had time to really look into it, if you want to close this feel free. And if you don't get round to implementing something based on `compiletest-rs` I might get time to look into it in the next few weeks.

iliekturtles:
I'll leave the PR open for the moment, but will plan to cleanup the patch I posted the other day to use `compiletest-rs`.

iliekturtles:
I started adding `compiletest` tests last week, but E0464 issues, even with the `clean_rmeta` work-around, seem very common so I'm less enthusiastic about this option and leaning back towards using a workspace.

iliekturtles:
Decided to go with the separate test crate: d0efe67cd245d0e4baf2ecd893a6c1262db7f48a

Issue #52, opened 2018-01-20
Improve compile times


iliekturtles:
Compiling `uom` with all features enabled takes a significant amount of time. `-Ztime-passes` shows that the issue is in privacy checking:

```sh
$ RUSTFLAGS="-Ztime-passes" cargo +nightly test --no-run
...
time: 172.929; rss: 507MB	privacy checking
...
```

Timing for this pass increases linearly with the number of underlying storage types enabled. Enabling features for underlying storage types adds `impl` blocks and type aliases for existing types. No new types are added. Additionally, all `struct`s and `trait`s are `pub` in `uom`.

The graph below shows compile times, as reported by `cargo` for `test --no-run` in blue and `build` in red:
![image](https://user-images.githubusercontent.com/5081378/35186824-7d88eecc-fde8-11e7-9d11-bb4244fae8f6.png)

Test | Build | Features
-- | -- | --
15.74 | 5.5 | f64
20.4 | 5.43 | f32,f64
38.84 | 6.24 | bigrational,f32,f64
37.75 | 7.4 | rational64,bigrational,f32,f64
48.28 | 8.26 | rational32,rational64,bigrational,f32,f64
55.52 | 9.26 | rational,rational32,rational64,bigrational,f32,f64
67.57 | 10.89 | biguint,rational,rational32,rational64,bigrational,f32,f64
96.81 | 12.5 | bigint,biguint,rational,rational32,rational64,bigrational,f32,f64
94.53 | 15.21 | i64,bigint,biguint,rational,rational32,rational64,bigrational,f32,f64
106.92 | 16.4 | i32,i64,bigint,biguint,rational,rational32,rational64,bigrational,f32,f64
116.34 | 18.17 | i16,i32,i64,bigint,biguint,rational,rational32,rational64,bigrational,f32,f64
131.29 | 20.2 | i8,i16,i32,i64,bigint,biguint,rational,rational32,rational64,bigrational,f32,f64
144.3 | 22.19 | isize,i8,i16,i32,i64,bigint,biguint,rational,rational32,rational64,bigrational,f32,f64
137.48 | 24.67 | u64,isize,i8,i16,i32,i64,bigint,biguint,rational,rational32,rational64,bigrational,f32,f64
158.4 | 26.93 | u32,u64,isize,i8,i16,i32,i64,bigint,biguint,rational,rational32,rational64,bigrational,f32,f64
[...]

PR #73, opened 2018-04-13
Add display tools for quantities.

### LINKED ISSUE #13: Implement Display, LowerExp, UpperExp, ...
https://doc.rust-lang.org/core/fmt/trait.Binary.html
https://doc.rust-lang.org/core/fmt/trait.Display.html
https://doc.rust-lang.org/core/fmt/trait.LowerExp.html
https://doc.rust-lang.org/core/fmt/trait.LowerHex.html
https://doc.rust-lang.org/core/fmt/trait.Octal.html
https://doc.rust-lang.org/core/fmt/trait.UpperExp.html
https://doc.rust-lang.org/core/fmt/trait.UpperHex.html

Is there a way to use the [fmt](https://doc.rust-lang.org/core/fmt/) syntax or flags to include an implicit unit conversion, unit abbreviation, or unit label?

```rust
// What replaces `...` to allow the following outputs: 10 m, 10 meters, 32.8084 ft, 32.8084 feet
println!("{...}", new Length(10, meter));
```

 - [ ] `Quantity`
 - [ ] `$quantities` (Dimension)
 - [ ] `BaseUnits`
 - [ ] `$unit` (Measurement units)

Aehmlo:
I'm not completely satisfied with this solution, and I welcome feedback, but it works.

Resolves #13.

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -1197,6 +1197,144 @@ macro_rules! system {
             }
         }
 
+        /// Contains formatting tools for displaying quantities in this system.
+        pub mod fmt {
+            use $crate::lib::fmt;
+            use super::{Dimension, Quantity, Unit, Units, from_base};
+            use $crate::num::Num;
+            use $crate::Conversion;
+
+            /// Specifies the display style to use.
+            #[derive(Clone, Copy, Debug)]
+            #[allow(dead_code)] // We don't use this internally.
+            pub enum DisplayStyle {
+                /// Display the value and a unit abbreviation, e.g. "1.0 m", "327 s".
+                Abbreviation,
+                /// Display the value and full unit name (pluralized as appropriate),
+                /// e.g. "1 kilogram", "756 feet".
+                Description,
+            }
+
+            #[doc(hidden)]
+            #[allow(dead_code)] // We don't use this internally.
+            pub struct QuantityArguments<T, D, U, V>
+            where
+                T: Unit,
+                D: Dimension + ?Sized,
+                U: Units<V> + ?Sized,
+                V: Num + Conversion<V>,
+            {
+                arguments: Arguments<T>,
+                quantity: Quantity<D, U, V>,
+            }
+
+            /// Provides an interface to specify a display style and unit.
+            ///
+            /// # Usage
+            /// ## Indirect style
+            /// ```
+            /// # use uom::si::Unit;
+            /// # use uom::si::f32::*;
+            /// # use uom::si::length::{centimeter, meter};
+            /// # use uom::si::fmt::Arguments;
+            /// # use uom::si::fmt::DisplayStyle::*;
+            /// let length = Length::new::<meter>(1.0);
+            /// let a = Arguments::new(centimeter, Description);
+            /// assert_eq!("100 centimeters", format!("{}", a.with(length)));
+            /// ```
+            /// ## Direct style
+            /// ```
+            /// # use uom::si::Unit;
+            /// # use uom::si::f32::*;
+            /// # use uom::si::length::{centimeter, meter};
+            /// # use uom::si::fmt::Arguments;
+            /// # use uom::si::fmt::DisplayStyle::*;
+            /// let length = Length::new::<meter>(1.0);
+            /// let a = Arguments::new_with(length, centimeter, Description);
+            /// assert_eq!("100 centimeters", format!("{}", a));
+            /// ```
+            #[allow(missing_debug_implementations)] // Prevent accidental direct use.
+            #[derive(Clone, Copy)]
+            pub struct Arguments<T>
+            where
+                T: Unit,
+            {
+                subunit: T,
+                style: DisplayStyle,
+            }
+
+            impl<T> Arguments<T>
+            where
+                T: Unit,
+            {
+                /// Creates an instance of `Arguments<T>` that can be used to format quantites
+                /// later.
+                ///
+                /// To format a quantity, use the `with` method on `Arguments<T>` to obtain
+                /// a displayable object.
+                #[allow(dead_code)] // We don't use this internally.
+                pub fn new(subunit: T, style: DisplayStyle) -> Self {
+                    Self { subunit, style }
+                }
+
+                /// Creates an immediately-displayable object that displays the passed quantity.
+                #[allow(dead_code)] // We don't use this internally.
+                pub fn new_with<D, U, V>(quantity: Quantity<D, U, V>, subunit: T, style: DisplayStyle) -> QuantityArguments<T, D, U, V>
+                where
+                    D: Dimension + ?Sized,
+                    U: Units<V> + ?Sized,
+                    V: Num + Conversion<V>,
+                {
+                    let a = Self::new(subunit, style);
+                    a.with(quantity)
+                }
+
+                /// Specifies a quantity to display.
+                #[allow(dead_code)]
+                pub fn with<D, U, V>(self, quantity: Quantity<D, U, V>) -> QuantityArguments<T, D, U, V>
+                where
+                    D: Dimension + ?Sized,
+                    U: Units<V> + ?Sized,
+                    V: Num + Conversion<V>,
+                {
+                    QuantityArguments {
+                        arguments: self,
+                        quantity,
+                    }
+                }
+            }
+
+            macro_rules! display_arguments {
+                ($style:ident) => {
+                    impl<T, D, U, V> fmt::$style for QuantityArguments<T, D, U, V>
+                    where
+                        T: Unit + Conversion<V, T = V::T>,
+                        D: Dimension + ?Sized,
+                        U: Units<V> + ?Sized,
+                        V: Num + Conversion<V> + fmt::$style,
+                    {
+                        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                            let value = from_base::<D, U, V, T>(&self.quantity.value);
+                            value.fmt(f)?;
+                            write!(f, " {}", match self.arguments.style {
+                                DisplayStyle::Abbreviation => T::abbreviation(),
+                                DisplayStyle::Description => if value.is_one() { T::singular() } else { T::plural() },
+                            })
+                        }
+                    }
+                };
+            }
+
+            display_arguments!(Binary);
+            display_arguments!(Debug);
+            display_arguments!(Display);
+            display_arguments!(LowerExp);
+            display_arguments!(LowerHex);
+            display_arguments!(Octal);
+            display_arguments!(UpperExp);
+            display_arguments!(UpperHex);
+        }
+
         /// Macro to implement [`quantity`](si/struct.Quantity.html) type aliases for a specific
         /// [system of units][units] and value storage type.
         ///


iliekturtles:
Thanks! First thing I notice is that the new `fmt` mod should be (at least partially) generated by the `system!` macro. Right now formatting only works for the `si` system when it should work for any system. Once this is done tests can be moved into the top-level `tests` module. I'll review in more detail this weekend.

src/si/fmt.rs
```diff
@@ -0,0 +1,133 @@
...
+display_arguments!(LowerExp, "{:e}");
+display_arguments!(LowerHex, "{:x}");
+display_arguments!(Octal, "{:o}");
+display_arguments!(UpperExp, "{:E}");
+display_arguments!(UpperHex, "{:X}");
+
+#[cfg(test)]
+mod tests {
+
+    use si::fmt::{Arguments, DisplayStyle::*};
```
iliekturtles: `uom` still supports `rustc` back to 1.20.0 which doesn't support nested imports.

src/si/fmt.rs
```diff
@@ -0,0 +1,133 @@
...
+        impl<T, D, U, V> fmt::$name for QuantityArguments<T, D, U, V>
+        where
+            T: Unit + Conversion<V>,
+            D: Dimension + ?Sized,
+            U: Units<V> + ?Sized,
+            V: Num + Conversion<V> + fmt::$name + Clone,
+        {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                let value = self.quantity.get::<T>(self.arguments.subunit);
+                write!(f, $pat, value)?;
```
iliekturtles: Instead of passing `$pat` in via macro parameter call the value's `fmt::$name` implementation: `value.fmt(f)` (or, if explicit call syntax is needed: `<V as ::lib::fmt::$name>::fmt(v, f)`). See the `Debug` impl for `Quantity` for an example.

src/si/fmt.rs
```diff
@@ -0,0 +1,133 @@
...
+
+}
+
+#[doc(hidden)]
+pub struct QuantityArguments<T: Unit, D: Dimension + ?Sized, U: Units<V> + ?Sized, V: Num + Conversion<V>> {
+    arguments: Arguments<T>,
+    quantity: Quantity<D, U, V>,
+}
+
+trait Convert<V> { // Use this get<N> instead of the (private) get<N> implemented per-unit.
```
iliekturtles: Would like to remove this trait. Calling the `from_base` directly should work.

src/system.rs
```diff
@@ -1071,6 +1071,101 @@ macro_rules! system {
...
+                /// Display the value and full unit name (pluralized as appropriate), e.g. "1 kilogram", "756 feet".
+                Description,
+            }
+
+            /// Provides an interface to specify a display style and unit.
+            ///
+            /// # Usage
+            /// ```
+            /// # use uom::si::Unit;
+            /// # use uom::si::{f32::*, length::{centimeter, meter}, fmt::{Arguments, DisplayStyle::*}};
```
iliekturtles: Remove nested imports. I'd like to get tests running for `rustc` 1.20.0 again.

src/system.rs
```diff
@@ -1071,6 +1071,101 @@ macro_rules! system {
...
+            /// ```
+            #[allow(missing_debug_implementations)] // Prevent accidental direct use.
+            #[derive(Clone)]
+            pub struct Arguments<T: Unit> {
+                subunit: T,
+                style: DisplayStyle,
+            }
+
+            impl<T: Unit> Arguments<T> {
+
```
iliekturtles: Remove blank line.

src/system.rs
```diff
@@ -1071,6 +1071,101 @@ macro_rules! system {
...
+                    D: Dimension + ?Sized,
+                    U: Units<V> + ?Sized,
+                    V: Num + Conversion<V>,
+                {
+                    QuantityArguments {
+                        arguments: self.clone(), // Allow use of this format style with multiple quantities
+                        quantity,
+                    }
+                }
+
```
iliekturtles: Remove blank line.

src/system.rs
```diff
@@ -1071,6 +1071,101 @@ macro_rules! system {
...
+            display_arguments!(Binary);
+            display_arguments!(Debug);
+            display_arguments!(Display);
+            display_arguments!(LowerExp);
+            display_arguments!(LowerHex);
+            display_arguments!(Octal);
+            display_arguments!(UpperExp);
+            display_arguments!(UpperHex);
+
```
iliekturtles: Remove blank line.

src/system.rs
```diff
@@ -1071,6 +1071,101 @@ macro_rules! system {
...
+            /// # Usage
+            /// ```
+            /// # use uom::si::Unit;
+            /// # use uom::si::{f32::*, length::{centimeter, meter}, fmt::{Arguments, DisplayStyle::*}};
+            /// let length = Length::new::<meter>(1.0);
+            /// let a = Arguments::new(centimeter, Description);
+            /// assert_eq!("100 centimeters", format!("{}", a.with(length)));
+            /// ```
+            #[allow(missing_debug_implementations)] // Prevent accidental direct use.
+            #[derive(Clone)]
```
iliekturtles: Why not `Copy`?

src/system.rs
```diff
@@ -1071,6 +1071,101 @@ macro_rules! system {
...
+
+            impl<T: Unit> Arguments<T> {
+
+                /// Creates a new `Argument` for displaying a quantity in the specified units and style.
+                pub fn new(subunit: T, style: DisplayStyle) -> Self {
+                    Self { subunit, style }
+                }
+
+                /// Specifies a quantity to display.
+                pub fn with<D, U, V>(&self, quantity: Quantity<D, U, V>) -> QuantityArguments<T, D, U, V>
```
iliekturtles: If `Arguments` was `Copy` I think this would be changed to taking self by value and not calling `clone()` down below.

src/system.rs
```diff
@@ -1071,6 +1071,101 @@ macro_rules! system {
...
+            }
+
+            macro_rules! display_arguments {
+                ($style:ident) => {
+                    impl<T, D, U, V> fmt::$style for QuantityArguments<T, D, U, V>
+                    where
+                        T: Unit + Conversion<V, T = V::T>,
+                        D: Dimension + ?Sized,
+                        U: Units<V> + ?Sized,
+                        V: Num + Conversion<V> + fmt::$style + Clone,
```
iliekturtles: Is `Clone` necessary?

src/system.rs
```diff
@@ -1071,6 +1071,101 @@ macro_rules! system {
...
+                    {
+                        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                            let value = from_base::<D, U, V, T>(&self.quantity.value);
+                            value.fmt(f).and_then(|_| {
```
iliekturtles: I think I liked the original code that used `?` better, e.g.

```rust
value.fmt(f)?;
write!(f, " {}", ...)
```

Aehmlo:
Well, it wasn't "the next few days" as I was hoping, but here we are.

I've fixed the above issues and added a mechanism for using `Arguments` directly and indirectly. There are two traits, `Direct` and `Indirect`, that the user can bring into scope in order to choose the behavior they want to use. How does this solution look to you?

src/system.rs
```diff
@@ -1197,6 +1197,162 @@ macro_rules! system {
...
+            /// # use uom::si::fmt::Arguments;
+            /// # use uom::si::fmt::Direct;
+            /// # use uom::si::fmt::DisplayStyle::*;
+            /// let length = Length::new::<meter>(1.0);
+            /// let a = Arguments::new(length, centimeter, Description);
+            /// assert_eq!("100 centimeters", format!("{}", a));
+            /// ```
+            #[allow(missing_debug_implementations)] // Prevent accidental direct use.
+            #[derive(Clone, Copy)]
+            pub struct Arguments<T: Unit> {
```
iliekturtles: Move generics bound for `T` into a where clause.

src/system.rs
```diff
@@ -1197,6 +1197,162 @@ macro_rules! system {
...
+                /// later.
+                ///
+                /// To format a quantity, use the `with` method on `Arguments<T>` to obtain
+                /// a displayable object.
+                fn new(subunit: T, style: DisplayStyle) -> Arguments<T>;
+            }
+
+            /// Allows for direct and immediate formatting of quantities in the specified style
+            /// and unit.
+            pub trait Direct<T>
```
iliekturtles: I'm thinking that using traits to overload `new()` adds too much complexity and just having two methods: `new(...) -> Arguments` and `new_with(...) -> QuantityArguments` makes things easier to use. With traits the trait needs to be in scope  and if you happen to have both in scope you need to use fully qualified method call syntax.

src/system.rs
```diff
@@ -1197,6 +1197,162 @@ macro_rules! system {
...
+            where
+                T: Unit {
+                fn new<D, U, V>(quantity: Quantity<D, U, V>, subunit: T, style: DisplayStyle) -> QuantityArguments<T, D, U, V>
+                where
+                    T: Unit,
+                    D: Dimension + ?Sized,
+                    U: Units<V> + ?Sized,
+                    V: Num + Conversion<V>,
+                    {
+                        let a: Arguments<T> = <Self as Indirect<T>>::new(subunit, style);
```
iliekturtles: Directly constructor the result? `QuantityArguments { arguments: Arguments { ... }, quantity }`

src/system.rs
```diff
@@ -1197,6 +1197,162 @@ macro_rules! system {
...
+            impl<T> Indirect<T> for Arguments<T>
+            where
+                T: Unit {
+                /// Creates a new `Argument` for displaying a quantity in the specified units and style.
+                fn new(subunit: T, style: DisplayStyle) -> Self {
+                    Self { subunit, style }
+                }
+            }
+
+            impl<T: Unit> Arguments<T> {
```
iliekturtles: Move generics bound for `T` into a where clause.

src/system.rs
```diff
@@ -1197,6 +1197,162 @@ macro_rules! system {
...
+                    QuantityArguments {
+                        arguments: self,
+                        quantity,
+                    }
+                }
+            }
+
+            #[doc(hidden)]
+            #[allow(dead_code)] // We don't use this internally.
+            pub struct QuantityArguments<T: Unit, D: Dimension + ?Sized, U: Units<V> + ?Sized, V: Num + Conversion<V>> {
```
iliekturtles: Move generics bounds into a where clause.

src/system.rs
```diff
@@ -1197,6 +1197,162 @@ macro_rules! system {
...
+                        D: Dimension + ?Sized,
+                        U: Units<V> + ?Sized,
+                        V: Num + Conversion<V> + fmt::$style,
+                    {
+                        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                            let value = from_base::<D, U, V, T>(&self.quantity.value);
+                            value.fmt(f)?;
+                            write!(f, " {}", match self.arguments.style {
+                                DisplayStyle::Abbreviation => T::abbreviation(),
+                                DisplayStyle::Description => if value.is_one() {T::singular()} else {T::plural()},
```
iliekturtles: Put spaces on each side of braces: `... else { T::plural() },`

src/system.rs
Aehmlo: I *believe* the compiler should inline this anyway, and it makes me uneasy to copy-paste code where I could just have a zero-cost call into another entry point. It's unlikely that this struct will need to change, but it's not impossible.

Aehmlo:
I've refactored this to simply use `new` and `new_with`; I agree that letting a little bit of the backing  hierarchy leak is better than requiring the user to manage trait import. A couple of lines break the 100-column rule, but I've left them for now, since I don't think they'd really benefit from reworking. I also haven't added tests outside of the doc tests; if you feel that more are needed, I'm happy to do so.

Aehmlo:
It looks to me that the build error is #52 again (although it's somewhat strange to see that it only happens on 1.20.0).

iliekturtles:
I think performance improvements since 1.20.0 allow newer versions of rustc to complete in less than 10 minutes.

iliekturtles:
Turns out we weren't as close as I thought at first. After reviewing in more depth I found a number of needed changes. I started making changes in my `dev-display` branch before I realized the full scope. I suggest squashing my commit into yours and then making the rest of the changes.

- Remove all the `#[allow(dead_code)]` attributes and add tests to `tests.rs` that go into more depth than the doc tests do. Check `Abbreviation` and `Description`. Check that other [format flags](https://doc.rust-lang.org/stable/std/fmt/index.html#syntax) (fill, align, sign, ...) work when used. Add a test for each trait to ensure they're being implemented correctly and pass through to the underlying type's implementation.
- I moved `QuantityArguments` below `Arguments`.
- Doc tests need `cfg_attr`s to only run if `f32` and `si` features are both defined.
- Add a `compile_fail` test to show/ensure that `Arguments` doesn't impl `Debug` (do for `Display` instead?).
- To be consistent with other code I renamed the `T` generic parameter to `N`.
- I renamed "subunit" to "unit."
- Any public methods that have a parameter of type `N` need to be moved into the `quantity!` macro. Currently you can do something like `Arguments::new_with(length, seconds, Description)` and use a `Quantity` and unit that don't match.
  - This also brings up an issue about how to print values for quantities that are not explicitly defined. I'm inclined to punt on this one for the moment.
  - Add a `compile_fail` test to show that mixing quantities and units wont work.

iliekturtles:
I don't think so, but can you elaborate some more? In regards to moving code into the `quantity!` macro it's because of the need to use a type's specific `Unit` trait. e.g. you shouldn't be able to use `seconds` as the unit for a quantity of `length`.

Aehmlo:
You're right, we can't constrain using `Dimension` as-is because `Unit`s don't contain any information about dimensions. I assume this is intentional, but it seems to me that if we were to use something like an associated type to embed dimensional information into the unit, we would have a nice solution here. Then we could do something like `where N: Unit<Dimension = D>` and eliminate the aforementioned issues.

It's very likely that there's some subtlety I'm missing, or that I'm just not thinking hard enough (it's pretty early for this), but this does seem to be a good solution to me that has some nice unit-safety benefits.

I investigated this glancingly, but I saw some things that might be problematic with `Sized` bounds. This may very well be impossible with the current architecture.

Aehmlo:
I'm going to go ahead and close this until I'm able to pick it back up again.

iliekturtles:
Just pushed the [work](https://github.com/iliekturtles/uom/tree/dev-display) I've put in so far. Did you want to take these changes and re-open this PR or review what I've done and I can complete the changes?

 * [ ] Review generic parameters.
 * [ ] `pub(crate)` necessary?
 * [ ] Finish comments.
   * [ ] Move examples to `format_args` and `to_format_args`?
 * [ ] Correct tests
   * [ ] `Clone`/`Copy` not derived correctly. Need custom implementations that ignore `Dimension`.
   * [ ] Assert tests for `DisplayStyle`?
   * [ ] Add tests for each trait (`Binary`, `Debug`, `Display`, ...) (Single test that runs through all the traits at once or separate tests?)
   * [ ] Add round trip test `Display` -> `FromStr`.

PR #76, opened 2018-04-17
Fix current documentation copy-paste error.

Aehmlo:
`electric_current.rs` currently has a minor [documentation error](https://github.com/iliekturtles/uom/blob/7d7cab0e3e35fe38d8eba5d2f6d56f74103b9da2/src/si/electric_current.rs#L6) stemming from a copy-paste that wasn't properly edited. This is a very minor thing, but obviously worth fixing.


iliekturtles:
Good catch. Thanks!

Issue #72, opened 2018-04-13
Volume rate quantity


iliekturtles:
Add volume rate quantity and associated units.

https://www.nist.gov/physical-measurement-laboratory/nist-guide-si-appendix-b9#VOLUMEtime

Issue #71, opened 2018-04-13
Mass rate quantity


iliekturtles:
Add mass rate quantity and associated units.

https://www.nist.gov/physical-measurement-laboratory/nist-guide-si-appendix-b9#MASStime

PR #79, opened 2018-04-21
Add hydrodynamics-related properties.

### LINKED ISSUE #68: Density quantity
Add density quantity and associated units.

https://www.nist.gov/pml/nist-guide-si-appendix-b9-factors-units-listed-kind-quantity-or-field-science#ENERGY

Aehmlo:
This closes #68, #70, #71, and #72.


iliekturtles:
Thanks for another huge PR! I didn't get to pressure and density yet.

- Mass rate
  - Wrap at 100 characters.
  - Add tests for `carat_per_second` through `ton_per_second`. I don't think the `pound_per_minute`/`_hour` conversion factors have enough precision. Add another generic parameter to the test function to accept the time unit.
  - Add at least `kilogram_per_minute`/`_hour`. Consider a few more like `gram_per_minute` or `_day` suffixes for got kilograms and pounds. The full prefix x prefix matrix isn't necessary but I'd like to see a few of the more common combinations.

iliekturtles:
- Pressure
  - Change `centimeter_of_mercury_4` to "`@centimeter_of_water_4: ...; "cm H₂O (4  °C)", "centimeter of water (4 °C)", "centimeters of water (4 °C)";`
  - Add tests for non-SI units where the force unit exists.
- Density
  - Some of the `<sup>` tags got messed up with `m³`.
  - `grain_per_gallon` conversion factor should be `E-2`.
  - Add test for non-SI units where the mass and volume units exist.

Aehmlo:
I've fixed these, and tests for these quantities are now much more robust.

Aehmlo:
My bad — used `--onto` incorrectly and pushed before realizing it!

PR #83, opened 2018-05-01
Basic rustfmt setup

### LINKED ISSUE #80: Setup rustfmt
Currently `uom` tries to use most `rustfmt` styles options but doesn't actually use `rustfmt` because `rustfmt` overrides a couple manual formatting choices and there is no way to skip these currently (the `#[cfg(...)]compile_error!(...);` lines in `src/lib.rs` mostly).

- `rustfmt_skip` doesn't seem to work.
- `file-lines` doesn't have a negative option to only exclude certain lines.

Aehmlo:
This should resolve #80 when we're done. This isn't the ideal solution, since it has to just skip a large chunk of `tests.rs`, but I think it's better than nothing, and it'll certainly cut down on the amount of "wrap at 100 characters" comments in your review process. :)

There were a couple of suggestions that `rustfmt` made that I liked, so I kept them even though I'm sure they were conscious decisions on your part. Of course, I'm willing to discuss these further.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,10 +1,10 @@
 //! Units of measurement is a crate that does automatic type-safe zero-cost
-//! [dimensional analysis][analysis]. You can create your own systems or use the pre-built
-//! [International System of Units][si] (SI) which is based on the
-//! [International System of Quantities][isq] (ISQ) and includes numerous [quantities][quantity]
-//! (length, mass, time, ...) with conversion factors for even more numerous
-//! [measurement units][measurement] (meter, kilometer, foot, mile, ...). No more crashing your
-//! [climate orbiter][orbiter]!
+//! [dimensional analysis][analysis]. You can create your own systems or use
+//! the pre-built [International System of Units][si] (SI) which is based on the
+//! [International System of Quantities][isq] (ISQ) and includes numerous
+//! [quantities][quantity] (length, mass, time, ...) with conversion factors
+//! for even more numerous [measurement units][measurement] (meter, kilometer,
+//! foot, mile, ...). No more crashing your [climate orbiter][orbiter]!
 //!
 //! [analysis]: https://en.wikipedia.org/wiki/Dimensional_analysis
 //! [si]: http://jcgm.bipm.org/vim/en/1.16.html
@@ -27,8 +27,8 @@
 //! extern crate uom;
 //! ```
 //!
-//! The simple example below shows how to use quantities and units as well as how `uom` stops
-//! invalid operations:
+//! The simple example below shows how to use quantities and units as well as
+//! how `uom` stops invalid operations:
 //!
 #![cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
 #![cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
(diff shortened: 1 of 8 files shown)


Aehmlo:
I think we would want to use a `rustfmt.toml` like the following (currently requires nightly), probably. I didn't look through all the possible parameters, though.
```toml
format_strings = true
error_on_unformatted = true
```

src/tests.rs
```diff
@@ -377,6 +378,7 @@ mod quantity_macro {
     }
 }
 
+#[cfg_attr(rustfmt, rustfmt_skip)] // rustfmt isn't currently flexible enough for us to do much else here
```
iliekturtles: This attribute wasn't necessary when I ran `cargo fmt`.

src/lib.rs
```diff
@@ -146,31 +146,31 @@
...
+#![warn(trivial_numeric_casts)]
+#![warn(unused_extern_crates)]
+#![warn(unused_import_braces)]
+#![warn(unused_qualifications)]
+#![warn(unused_results)]
 // Clippy lints.
 #![cfg_attr(feature = "cargo-clippy", allow(inline_always))]
 
 // Fail to compile if no underlying storage type features are specified.
+#[cfg_attr(rustfmt, rustfmt_skip)]
```
iliekturtles: Glad you got this one working! Was the biggest blocker last time I tried.

src/lib.rs
```diff
@@ -146,31 +146,31 @@
 
 // Compile with `no_std` when the `std` feature is not specified.
 #![cfg_attr(not(feature = "std"), no_std)]
-
```
iliekturtles: I would have liked to maintain these blank lines, but that doesn't seem possible.

iliekturtles:
https://crate-ci.github.io/pr/rustfmt.html

Aehmlo:
Is it useful to remove the unstable (i.e. nightly-only) features from our setup until they're stabilized, or do we want to do formatting with nightly? The commit I've pushed above contains the results of running `rustfmt` on the nightly channel, but I understand if we want to use something more stable than that.

I do like the idea of enforcing style on CI builds, but I'll leave that for you to wade through. 😃

Side note: there's a point in `.travis.yml` that reads `cargoo` instead of `cargo`.

iliekturtles:
`trailing_comma = "Always"` isn't working like I expected based on the name. Please remove that one.

I also get weird output on both stable and nightly where `rustfmt` is inserting additional comments. Issue seems to be caused by `wrap_comments = true` so remove that one as well. I'll look into submitting an issue upstream.

```diff
diff --git a/src/lib.rs b/src/lib.rs
index f1c54a5..f8d00b1 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,7 +24,9 @@
 //! and this to your crate root:
 //!
 //! ```rust
+//!
 //! extern crate uom;
+//! }
 //! ```
 //!
 //! The simple example below shows how to use quantities and units as well as
```

I would like to use stable... however stable currently crashes so I suggest we use beta which will be stable in less than 5 weeks and supports all the options we have enabled. I'll also see about preparing a change to verify formatting during CI to be merged shortly after 1.27 is released.

```sh
~/Source/uom (aehmlo-rustfmt *$)$ cargo fmt
thread 'main' panicked at 'internal error: entered unreachable code', tools\rustfmt\src\macros.rs:701:18
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Aehmlo:
I think we should be good now.

iliekturtles:
Seeing a number of new changes that are wrapping comments at 80 characters. Can you revert those and rename `rustfmt.toml` to `.rustfmt.toml`?

Aehmlo:
I’ve been just pushing the code resulting from running `cargo fmt` so that it doesn’t fail future CI style checking, although I suppose if `wrap_comments` is misbehaving that there’s not much to do about that for now. I’ll fix that manually and we can incorporate the real fix later when `rustfmt` is patched.

iliekturtles:
I was unable to reproduce `cargo [|+beta|+nightly] fmt` changing the comment formatting. Perhaps it got run once with no `rustfmt.toml` which I think would default to 80 characters.

iliekturtles:
Today's 1.26.1 release comes with a new `rustfmt` that fixes the crash seen in the previous stable. I pushed a minimal set of changes as well as a validation step in Travis CI if you want to review: https://github.com/iliekturtles/uom/commit/366b90e35368efe9bb3d502ef2cb7055b8f823a1. I'll plan on merging Thursday if nothing comes up.

Aehmlo:
LGTM. Glad that 1.26.1 fixed the crash!

Aehmlo:
Closed in 366b90e.

PR #87, opened 2018-05-20
Removed unused `_unit: N` parameters and require turbofish.

iliekturtles:
`_unit` parameter already removed from `new()` and provides no benefit
or utility as only the type is important and must be statically known at
compile time.

---

Breaking change. I'm reviewing further breaking changes as currently there is no way for unit selection to be done based on user input.


iliekturtles:
Decided to merge these changes and opened #91 to investigate how to best handle dynamic unit selection.

PR #90, opened 2018-06-23
Implement FromStr for Quantity.

### LINKED ISSUE #14: Implement FromStr
https://doc.rust-lang.org/core/str/trait.FromStr.html

 - [ ] `Quantity`
 - [ ] `$unit` (Measurement units)

Aehmlo:
I'm really not a huge fan of this implementation, but it does work. My greatest concern is definitely `where $( super::$unit: $crate::Conversion<V, T = V::T>, )+`, which just feels dirty (and causes the compiler to think that a pattern is unreachable just because there are so many options). I think that we could probably avoid this by embedding the code inside a `storage_type!` block somewhere, but when I tried that, I ran into a whole slew of problems. Maybe when I find the mental fortitude, I'll try again.

There are also no tests here; after wrestling with macro errors for a few hours, I just couldn't summon up the energy to parse `tests.rs` (which is getting to be pretty dang big!). I'd really appreciate some help writing these.

This approach enables e.g. `"1.4 fm".parse::<Length>()`, while `"1.4 fs".parse::<Length>()` will error.

I wrote some docs, but they can definitely be improved (and I'd be happy to fix them up once we've finalized an implementation, but I have a feeling we won't be sticking with this one).

When all is said and done, this will close #14! 🎉


Aehmlo:
I also want to make this a little more lenient, enabling users to use both abbreviations and full names, as well as making superscripts less finicky, allowing e.g. `<sup>2</sup>` and `^2` in addition to `²`.

src/quantity.rs
```diff
@@ -210,6 +210,54 @@ macro_rules! quantity {
             impl super::Conversion<V> for super::$unit {})+
         }
 
+        #[cfg(feature = "std")]
+        mod parse {
```
iliekturtles: `mod str` to match `FromStr`'s location?

src/quantity.rs
```diff
@@ -210,6 +210,54 @@ macro_rules! quantity {
...
+            use $crate::lib::str::FromStr;
+
+            /// Represents an error encountered while parsing a string into a `Quantity`.
+            #[derive(Clone, Copy, Debug)]
+            pub enum ParseError {
```
iliekturtles: This `enum` can be in `lib.rs` since it isn't system/quantity specific.

src/quantity.rs
```diff
@@ -210,6 +210,54 @@ macro_rules! quantity {
...
+                /// The abbreviation used wasn't found for this quantity.
+                UnknownAbbreviation,
+            }
+
+            impl<U, V> FromStr for super::$quantity<U, V>
+            where
+                U: super::super::Units<V> + ?Sized,
+                V: $crate::num::Num + $crate::Conversion<V> + FromStr,
+                $(
+                    super::$unit: $crate::Conversion<V, T = V::T>,
```
iliekturtles: I see what you mean about this being ugly. What is the error if removed?

src/quantity.rs
```diff
@@ -210,6 +210,54 @@ macro_rules! quantity {
...
+                )+
+            {
+                type Err = ParseError;
+                #[allow(unreachable_patterns)]
+                fn from_str(s: &str) -> Result<Self, Self::Err> {
```
iliekturtles: Will need to investigate this in more detail. May be something where this implementation can be accepted and then optimized later (finite state table vs. `match`?).

 * `contains` and `split` cause part of the `&str` to be iterated over multiple times. What about finding he index of the first space and then extracting the two sub-slices.
 * What if an abbreviation has spaces in it? Won't this only attempt to parse the first portion as the abbreviation?

iliekturtles:
I'm hesitant about using `regex` and `lazy_static` but would like to see what you have implemented if you've pushed the changes somewhere. My gut feeling is that implementing parsing as a state machine would be most efficient and could easily include rules such as accepting different forms of superscripts. This would require some compile-time code generation in excess of what macros could do however. (#60 will likely require additional compile-time code generation anyway)

src/quantity.rs
Aehmlo: I think that using [`splitn`](https://doc.rust-lang.org/std/primitive.str.html#method.splitn) will resolve these concerns.

src/quantity.rs
Aehmlo: One isolated error:
```
error[E0277]: the trait bound `si::volume_rate::register_ton_per_second: Conversion<V>` is not satisfied
   --> src/quantity.rs:307:49
    |
307 |                               $abbreviation => Ok(Self::new::<super::$unit>(value)),
    |                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Conversion<V>` is not implemented for `si::volume_rate::register_ton_per_second`
    |
   ::: src/si/volume_rate.rs:3:1
    |
3   | / quantity! {
4   | |     /// Volume rate (base unit cubic meter per second, m<sup>3</sup> · s<sup>-1</sup>).
5   | |     quantity: VolumeRate; "volume rate";
6   | |     /// Volume dimension, m<sup>3</sup> · s<sup>-1</sup>.
...   |
108 | |     }
109 | | }
    |
    = help: consider adding a `where si::volume_rate::register_ton_per_second: Conversion<V>` bound
note: required by `si::volume_rate::<impl si::Quantity<si::Dimension<L=typenum::PInt<typenum::UInt<typenum::UInt<typenum::UTerm, typenum::B1>, typenum::B1>>, M=typenum::Z0, T=typenum::NInt<typenum::UInt<typenum::UTerm, typenum::B1>>, Th=typenum::Z0, J=typenum::Z0, N=typenum::Z0, I=typenum::Z0> + 'static, U, V>>::new`
   --> src/quantity.rs:329:13
    |
329 | /             pub fn new<N>(v: V) -> Self
330 | |             where
331 | |                 N: Unit + $crate::Conversion<V, T = V::T>,
332 | |             {
...   |
337 | |                 }
338 | |             }
    | |_____________^
    |
   ::: src/si/volume_rate.rs:3:1
    |
3   | / quantity! {
4   | |     /// Volume rate (base unit cubic meter per second, m<sup>3</sup> · s<sup>-1</sup>).
5   | |     quantity: VolumeRate; "volume rate";
6   | |     /// Volume dimension, m<sup>3</sup> · s<sup>-1</sup>.
...   |
108 | |     }
109 | | }
```

[Full error](https://gist.github.com/Aehmlo/e6ef84a319bc23cc27ae417e8f403928) (huge due to macro invocations).

As I mentioned, I'm sure there are better ways/workarounds to do this.

src/quantity.rs
Aehmlo: I do like having it within `str`, although we could make a top-level `mod str`, I guess. What would you prefer?

Aehmlo:
I'll rework the smaller things above and rebase/push those changes this evening, and I'll also push branches for both the `regex` and `regex` + `lazy_static` approaches for you to investigate in the near future (not all implementation is where I want it yet).

I'll probably also write some isolated test cases and put them somewhere for later integration into the existing tests.

Aehmlo:
My new branches [`fromstr-regex`](https://github.com/Aehmlo/uom/tree/fromstr-regex) and [`fromstr-regex-static`](https://github.com/Aehmlo/uom/tree/fromstr-regex-static) add regex and cached regex support to the implementation, respectively. I've also pushed a revision of the current implementation with some of the changes from above. (An aside: I'd love to reduce allocations by one per invocation by not using `format!` directly, but this may be more trouble than it's worth.)

Aehmlo:
While I'm at it, [here are a handful of test cases I wrote up](https://gist.github.com/Aehmlo/5e93579004282dbdb3512905652cf897). They're far from comprehensive, but they should give an illustration of what sorts of things I'm aiming to support (these are all passing under my `fromstr-regex` branch).

iliekturtles:
* Why is `#[cfg(feature = "std")]` needed? Shouldn't parse be able to work in `no_std`?
 * Looking at how other parse error types are organized I suggest just moving the enum directly into the top-level and rename to `ParseQuantityError`.
   * Other error types use a pattern of `struct ParseXYZError { kind: XYZErrorKind } enum XYZErrorKind { ... }`. Why not use the same pattern? I haven't investigated enough to determine why it's done that way.
 * The `super::$unit: $crate::Conversion<...>` is required because `Conversion` is only implemented for the underlying storage types and not all `V`. Move the `FromStr` impl into a `storage_types!` invocation to get around this issue.

iliekturtles:
* I'm not inclined to include the regex implementations currently. They add a lot of complexity and requirements (regex reference, std requirement, allocations) in order to gain some flexibility in what can be successfully parsed as a `Quantity`. I think being able to round-trip `Display` and `FromStr` is sufficient at this point.
   * I still think a pre-computed finite state machine parser is the right way to go to support abbreviation/singular/plural as well other flexible rules like `^2`, `<sup>2</sup>`, or even re-parsing `Debug` output.
 * Why just match tab and 2+ spaces instead of using the `\s` (`\p{Whitespace}`) character class?

Aehmlo:
Answering in order:
* You're absolutely right; for whatever reason, I thought `FromStr` didn't exist in `core`.
* I can do this; I also have no idea why this pattern is so prevalent, but I guess the fact that it is suggests that we should follow suit.
* This is what I tried originally, but it ended up having strange errors; I will try again now that I've had some distance from it for a while. (I bet it works on the first try now.)
* 👍
* I matched tabs and spaces explicitly because `\s` matches both horizontal and vertical whitespace, and `regex` wraps greedily by default, and this combination meant that some strange things happened with newlines/carriage feeds in some cases. It was meant as a temporary workaround, and isn't relevant anymore, but that was the reasoning there.

I expect to make these changes tonight/tomorrow, and will push them ASAP thereafter.

Aehmlo:
Re: `enum` inside of `struct` for error handling: I believe that it's just for the sake of opacity. In this pattern, the underlying `enum` is private. In our case, I don't see any strong reason that we shouldn't let users match against the type of error, so I vote for keeping it as an `enum`.

Aehmlo:
As you'll notice, the code I just pushed will fail in CI. The error message is of the form:
```
error[E0119]: conflicting implementations of trait `std::str::FromStr` for type `si::Quantity<si::Dimension<T=typenum::NInt<typenum::UInt<typenum::UTerm, typenum::B1>>, Th=typenum::Z0, M=typenum::Z0, L=typenum::PInt<typenum::UInt<typenum::UInt<typenum::UTerm, typenum::B1>, typenum::B1>>, N=typenum::Z0, J=typenum::Z0, I=typenum::Z0> + 'static, _, _>`:
   --> src/quantity.rs:220:17
    |
220 |                   impl<U, V> FromStr for super::super::$quantity<U, V>
    |  _________________^
    | |_________________|
    | |
221 | |                 where
222 | |                     U: super::super::super::Units<V> + ?Sized,
223 | |                 {
...   |
237 | |                     }
238 | |                 }
    | |                 ^
    | |_________________|
    | |_________________first implementation here
    |                   conflicting implementation for `si::Quantity<si::Dimension<T=typenum::NInt<typenum::UInt<typenum::UTerm, typenum::B1>>, Th=typenum::Z0, M=typenum::Z0, L=typenum::PInt<typenum::UInt<typenum::UInt<typenum::UTerm, typenum::B1>, typenum::B1>>, N=typenum::Z0, J=typenum::Z0, I=typenum::Z0> + 'static, _, _>`
    |
   ::: src/si/volume_rate.rs:3:1
    |
3   | / quantity! {
4   | |     /// Volume rate (base unit cubic meter per second, m<sup>3</sup> · s<sup>-1</sup>).
5   | |     quantity: VolumeRate; "volume rate";
6   | |     /// Volume dimension, m<sup>3</sup> · s<sup>-1</sup>.
...   |
108 | |     }
109 | | }
```
There are 30 such errors (one for each quantity in `si`). I haven't dug into `storage_types!` enough to really grok what's going on here; I can't imagine it would be a conflict with the pointer-sized integers, right? Regardless, something's up here, although I think we'll be in the clear once we figure out what it is.

iliekturtles:
Remove the generic parameter `V` from your `impl`. The `storage_types!` macro introduces a local type alias `V`: `impl<U> FromStr for super::super::$quantity<U, V> { ... }`

Aehmlo:
Ah, that’s clever! I’m mobile right now, but I’ll fix that up in a couple of hours.

Aehmlo:
I've fixed the compiler errors, added support for underscores (apparently some storage types don't properly support underscores in literals?), and added a couple of test cases.

iliekturtles:
Excellent. Should have time to review later today!

src/lib.rs
```diff
@@ -442,3 +442,20 @@ storage_types! {
...
+    pub enum ParseQuantityError {
+        /// No separators (spaces) were encountered.
+        NoSeparator,
+        /// An error occured while parsing the value (first) portion of the string.
+        ///
+        /// Due to exhaustiveness and type system limitations, this variant does not encode
+        /// the underlying parse error.
+        ValueParseError,
+        /// The abbreviation used wasn't found for this quantity.
+        UnknownAbbreviation,
```
iliekturtles: Consider changing to `UnknownUnit` and note in the documentation that currently only the abbreviation is checked but future updates will include the description.

src/quantity.rs
```diff
@@ -210,6 +210,35 @@ macro_rules! quantity {
             impl super::Conversion<V> for super::$unit {})+
         }
 
+        mod str {
```
iliekturtles: Move down beneath the `impl<U, V> $quantity<U, V>` block.

src/quantity.rs
```diff
@@ -210,6 +210,35 @@ macro_rules! quantity {
...
+            storage_types! {
+                types: All;
```
iliekturtles: Line can be removed. No types declaration will expand to `types: All;`.

src/quantity.rs
```diff
@@ -210,6 +210,35 @@ macro_rules! quantity {
...
+
+                use $crate::lib::str::FromStr;
+                use $crate::str::ParseQuantityError::*;
+
+                impl<U> FromStr for super::super::$quantity<U, V>
+                where
+                    U: super::super::super::Units<V> + ?Sized,
+                {
+                    type Err = $crate::str::ParseQuantityError;
```
iliekturtles: Add a blank line below this one.

src/quantity.rs
```diff
@@ -210,6 +210,35 @@ macro_rules! quantity {
...
+                    #[allow(unreachable_patterns)]
```
iliekturtles: Still needed?

src/quantity.rs
```diff
@@ -210,6 +210,35 @@ macro_rules! quantity {
...
+                    fn from_str(s: &str) -> Result<Self, Self::Err> {
+                        let mut parts = s.splitn(2, ' ');
+                        let value = parts.next().unwrap();
+                        let value = value.replace('_', "").parse::<V>().map_err(|_| ValueParseError)?;
+                        let abbr = parts.next().ok_or(NoSeparator)?;
```
iliekturtles: Add a blank line below this one.

src/quantity.rs
```diff
@@ -210,6 +210,35 @@ macro_rules! quantity {
...
+                impl<U> FromStr for super::super::$quantity<U, V>
+                where
+                    U: super::super::super::Units<V> + ?Sized,
+                {
+                    type Err = $crate::str::ParseQuantityError;
+                    #[allow(unreachable_patterns)]
+                    fn from_str(s: &str) -> Result<Self, Self::Err> {
+                        let mut parts = s.splitn(2, ' ');
+                        let value = parts.next().unwrap();
+                        let value = value.replace('_', "").parse::<V>().map_err(|_| ValueParseError)?;
```
iliekturtles: * I'm hesitant about the `'_'` support as this differs from the standard library.
 * Just looking at the web view so I can't easily see the line length. May need to wrap if it exceeds 100.

```rust
let value = value.replace(...)
    .parse::<V>()
    .map_err()?;
```

src/quantity.rs
```diff
@@ -210,6 +210,35 @@ macro_rules! quantity {
...
+
+                impl<U> FromStr for super::super::$quantity<U, V>
+                where
+                    U: super::super::super::Units<V> + ?Sized,
+                {
+                    type Err = $crate::str::ParseQuantityError;
+                    #[allow(unreachable_patterns)]
+                    fn from_str(s: &str) -> Result<Self, Self::Err> {
+                        let mut parts = s.splitn(2, ' ');
+                        let value = parts.next().unwrap();
```
iliekturtles: I don't like the `unwrap()`, but I haven't thought of a better way to handle this since I'm pretty sure that `splitn` will always return at least one group.

src/quantity.rs
```diff
@@ -210,6 +210,35 @@ macro_rules! quantity {
...
+                        let value = value.replace('_', "").parse::<V>().map_err(|_| ValueParseError)?;
+                        let abbr = parts.next().ok_or(NoSeparator)?;
+                        match abbr.trim() {
+                            $(
```
iliekturtles: Join these three lines: `$($abbreviation => Ok(...),)+`

src/tests/quantity.rs
```diff
@@ -31,6 +31,19 @@ storage_types! {
         Test::assert_eq(&V::one(), &m1.get::<kilogram>());
     }
 
+    #[test]
+    fn from_str() {
```
iliekturtles: Add some negative tests:

 * No separator.
 * Invalid number that does not parse.
 * Invalid abbreviation (junk, abbreviation from different unit).
 * Valid abbreviation followed by extra: `1 m asdf`.

src/quantity.rs
Aehmlo: Yep, I believe it does. Short of using `collect` and subscripts, I don't think there's a way to express this without `unwrap`.

src/quantity.rs
Aehmlo: Let's go ahead and remove the underscore support and just have users strip them themselves (since the only case I can see where that'd be a big issue is if there's a UI involved, and in that case, input processing will already be more involved).

Aehmlo: (Removing this also brings the line back under 100 characters.)

src/quantity.rs
Aehmlo: Yes, which is somewhat strange (but I think probably just a limitation of the current macros implementation).

Aehmlo:
All right, the one thing I think we have left to address is the negative test cases; I'm not familiar with the test runner you're using, and so I implemented them using `assert_eq!` instead of `Test::assert_eq` due to some trait requirements I encountered. I can either revisit this later tonight or leave it to you to correct. Regardless, there are now negative tests that function as expected.

iliekturtles:
Using `assert_eq!` (or associated macros) is correct. `Test::assert_eq` is just https://github.com/iliekturtles/uom/blob/b22c6967bdb10bb29124b69e3c16a0805827ca16/src/tests/mod.rs#L53 and exists so that the same test code can work slightly differently depending on the underlying storage type. Floats do approximate equality while other types do exact equality.

I'll review all the changes later today hopefully.

src/lib.rs
```diff
@@ -442,3 +442,24 @@ storage_types! {
         }
     }
 }
+
+/// Shared tools for parsing quantities from strings.
+pub mod str {
+    /// Represents an error encountered while parsing a string into a `Quantity`.
+    #[derive(Clone, Copy, Debug, PartialEq)]
```
iliekturtles: Should this drop `Copy` and add `Eq`? Standard library error types use `#[derive(Debug, Clone, PartialEq, Eq)]`. Not sure why `Copy` is excluded. Perhaps only to make copying an error more explicit?

src/lib.rs
```diff
@@ -442,3 +442,24 @@ storage_types! {
         }
     }
 }
+
+/// Shared tools for parsing quantities from strings.
```
iliekturtles: `mod str` isn't just for parsing. `core::str`'s documentation is `String manipulation`, `std::str` is `Unicode string slices`. Consider changing to `Unicode string slice manipulation for quantities.`

src/lib.rs
```diff
@@ -442,3 +442,24 @@ storage_types! {
...
+pub mod str {
+    /// Represents an error encountered while parsing a string into a `Quantity`.
+    #[derive(Clone, Copy, Debug, PartialEq)]
+    pub enum ParseQuantityError {
+        /// No separators (spaces) were encountered.
+        NoSeparator,
```
iliekturtles: Add a blank line beneath each enum member.

Aehmlo:
> we have some erroneous duplicate abbreviations

This is at least a nice lint. 😄 Would you be open to another PR introducing a test that uses `#[deny(unreachable_patterns)]` (or something better) to ensure there are no duplicates in any abbreviations/descriptions/etc.?

I've resolved all the existing abbreviation conflicts in [dd0df84](https://github.com/Aehmlo/uom/commit/a4ba1609e6cbd485a34b2f3570895d2fedd0df84), but I'm not happy with all of the changes; if you could review them them and provide feedback, that'd be great.

iliekturtles:
Couple minor changes and then I believe we can merge.
 * Made notes header smaller.
 * Moved `use str::ParseQuantityError;` into root tests module.
 * Shortened string so that the line doesn't exceed 100 characters.

I would accept a test to `#[deny(unreachable_patterns)]` although I'm not completely convinced we should enforce it, I'm OK with it for the time being.

I also think the abbreviation changes are good enough for now. The mho symbol doesn't display properly in my editor or diff viewer, but works in the browser/terminal so I decided to leave it as-is for the moment.

```diff
diff --git a/src/lib.rs b/src/lib.rs
index f648002..8f17bdc 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -460,7 +460,7 @@ pub mod str {

         /// The unit used wasn't found for this quantity.
         ///
-        /// # Notes
+        /// ### Notes
         /// For now, only abbreviations are supported, so this error may be encountered even if the
         /// unit name (description) is correct.
         UnknownUnit,
diff --git a/src/tests/mod.rs b/src/tests/mod.rs
index c07fe52..0e6e8c0 100644
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -9,6 +9,7 @@ use num::{Float, FromPrimitive, One, Saturating, Signed, Zero};
 use quickcheck::TestResult;
 #[cfg(feature = "serde")]
 use serde_json;
+use str::ParseQuantityError;
 #[allow(unused_imports)]
 use typenum::{N1, P1, P2, P3, Z0};
 #[allow(unused_imports)]
diff --git a/src/tests/quantity.rs b/src/tests/quantity.rs
index ee614d8..d11f2d9 100644
--- a/src/tests/quantity.rs
+++ b/src/tests/quantity.rs
@@ -3,7 +3,6 @@
 #[cfg(feature = "autoconvert")]
 storage_types! {
     use tests::*;
-    use super::super::super::str::ParseQuantityError;

     mod f { Q!(tests, super::V); }
     mod k { Q!(tests, super::V, (kilometer, kilogram)); }
@@ -48,7 +47,7 @@ storage_types! {
         assert_eq!(&"10k m".parse::<k::Length>(), &Err(ParseQuantityError::ValueParseError));
         assert_eq!(&"1,000 km".parse::<k::Length>(), &Err(ParseQuantityError::ValueParseError));
         assert_eq!(&"10 s".parse::<k::Length>(), &Err(ParseQuantityError::UnknownUnit));
-        assert_eq!(&"10 kg // 10,000 g".parse::<k::Length>(), &Err(ParseQuantityError::UnknownUnit));
+        assert_eq!(&"10 kg //10,000 g".parse::<k::Length>(), &Err(ParseQuantityError::UnknownUnit));
     }

     quickcheck! {
```

Aehmlo:
I think we're finally done here. If this looks good to you, I'll go ahead and get back to work on ye olde `fmt` PR.

iliekturtles:
Merged! Thanks so much for all the effort!

I've recently put some effort into the `fmt` changes. Let me finish that or get it to a point where you can pick it up again.

PR #96, opened 2018-07-07
Allow for multiple quantities of the same dimension / implement temperature interval

iliekturtles:
Looking for review before merging into master. Note that Fahrenheit/Celsius/Rankine units will be done separately as part of #3.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -247,6 +247,68 @@ pub mod num {
     }
 }
 
+/// Primitive traits and types representing basic properties of types.
+pub mod marker {
+    /// Trait to denote that a quantity is able to be added with a quantity of the same dimensions.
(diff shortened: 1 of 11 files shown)


Aehmlo:
LGTM, apart from the code formatting and documentation of the marker traits (trivial). Elegant solution!

iliekturtles:
I ran out of time yesterday which is why the marker traits are still a mess. Another issue I ran into is that because specialization doesn't allow you to specialize both an associated type and a trait method at this point adding new overloads that change the associated type don't work. e.g. `ThermodynamicTemperature - ThermodynamicTemperature = TemperatureInterval` can't work currently. Since specialization is still unstable I've decided to punt for the moment and move on.

Aehmlo:
🎉

Issue #99, opened 2018-08-07
Update README regarding zero-cost guarantees


iliekturtles:
* rust-lang/rust#38269 is now fixed. Verify which rustc version it is in. Update readme with more details
 * Details about `autoconvert`
 * Add issue to create tests to verify zero-cost? Use `cargo asm`?

Thanks for the reminder @zesterer.

iliekturtles:
Zero-cost tests should be split into a new issue when this issue is worked on. Post v1.0.0 feature. Complete documentation for v1.0.0.

PR #103, opened 2018-08-27
Remove codegen comment

ThomasdenH:
There was a note about suboptimal codegeneration, but the issue seems to be resolved.


iliekturtles:
Thanks for the PR! #99 got added a few weeks ago to fix this as well as some other documentation issues and hasn't gotten any attention yet. Could you make the following changes and then I'll merge your PR:

 * `README.md` is also duplicated in `src/lib.rs`. Please make the same change there. (Any ideas to remove this duplication also welcome).
 * Could you find out version of `rustc` this fix was released in and note that.
 * Reference #99 in your commit message.

iliekturtles:
I just pushed an update to note that the codegen bug is fixed in `rustc` 1.25.0.

PR #120, opened 2019-03-25
Correct macros to generate valid 2015/2018 edition code.

### LINKED ISSUE #119: Doctest failure for custom quantity
I've created a custom quantities system and for some weird reasons doc example for the `quantity!` macro somehow breaks the documentation tests in my crate with the following error:

```
$ cargo test --all
..

failures:

---- <::uom::system::system macros> - IEC (line 803) stdout ----
error[E0432]: unresolved import `mks`
  --> doctest.rs:861:5
   |
59 |     Q!(mks, f32/*, (centimeter, gram, second)*/);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `mks` external crate
   |
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

thread '<::uom::system::system macros> - IEC (line 803)' panicked at 'couldn't compile the test', src/librustdoc/test.rs:351:13
```

Commenting [these lines](https://github.com/iliekturtles/uom/blob/54cc327ca2778c2c229c6c1c0c486d3ae4ddc973/src/system.rs#L1457-L1459) in the `uom` sources removes an error completely.

`rustc` version: 1.33.0
`uom` version:

```
[dependencies.uom]
version = "0.21.1"
default-features = false
features = ["autoconvert", "usize", "u64", "f64", "si", "std"]
```

I guess, doc test for exported `IEC!` macro breaks for some unknown reasons?

So, here is what makes this bug "weird": it is reproducible in the existing code, but I can't reproduce it with a new cargo project.
Unfortunately, existing code is not public (yet), so at the moment I'm failing to find the way of how to show it.

@iliekturtles, any thoughts, maybe?

iliekturtles:
Doc test on the `$quantities!` macro would fail with the error `no
``mks`` external crate` when the `system!` macro was invoked in a crate
using the 2018 edition. All invocations of the `$quantities!` macro are
corrected and a new test crate, `edition_check`, is added to validate
`uom` can be used in 2018 edition code. Resolves #119.

@svartalf if you could review that would be great.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,8 @@ matrix:
     script: |
       set -e
       cargo test --verbose --features "use_serde"
-      cargo test --manifest-path tests/feature_check/Cargo.toml --verbose
+      cargo test -p feature_check --verbose --features "use_serde"
+      cargo test -p edition_check --verbose --features "use_serde"
       cargo test --verbose --no-default-features --features "autoconvert f32 f64 si use_serde"
       cargo test --verbose --no-run --no-default-features --features "autoconvert usize u8 u16 u32 u64 isize i8 i16 i32 i64 bigint biguint rational rational32 rational64 bigrational f32 f64 std use_serde"
   - name: Rustfmt
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ maintenance = { status = "actively-developed" }
 [workspace]
 members = [
     "uom-macros",
+    "tests/edition_check",
     "tests/feature_check",
 ]
 
--- a/examples/mks.rs
+++ b/examples/mks.rs
@@ -81,9 +81,9 @@ system! {
 }
 
 mod f32 {
-    mod s {
-        pub use *;
+    mod mks {
+        pub use super::super::*;
     }
 
-    Q!(f32::s, f32);
+    Q!(self::mks, f32);
 }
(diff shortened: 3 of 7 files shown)


iliekturtles:
I ended up asking for a review too soon. I forgot that older versions don't support `crate` in paths. Trying to find a solution that supports old `rustc` versions.

iliekturtles:
@svartalf Ready for a review in my [dev-2018](https://github.com/iliekturtles/uom/tree/dev-2018) branch if you have some time. I'll try to merge in the next few days and push out another version to crates.io after that.

svartalf:
@iliekturtles what minimal rust version will be required by the "next" `uom` version?

If you're going to stick with `1.24` (same as now), I guess that your changes is the best option, since both `crate` and `$crate` are available only in `>=1.30`

 I've tested this branch with my #119 code also, works flawlessly :)

iliekturtles:
I plan to keep supporting 1.22.0 for compiling `uom` and 1.24.1 for full support for the foreseeable future. I didn't consider this requirement a big enough justification for a bump in the required version since there is an easy (although ugly) work-around.

Eventually the requirement will get bumped up. I'm doing some experimentation with proc-macros and how quantities track their units in the [dev-macros](https://github.com/iliekturtles/uom/tree/dev-macros) branch (very rough still). I'm contemplating calling the current `uom` 1.0 and if that branch works out releasing it as 2.0.

PR #116, opened 2019-03-21
Add liter-based volume units to si

Aehmlo:
This PR adds liter-based (*i.e.* `liter` and all prefixes) units to `si::volume`. The included tests test `liter` against `cubic_meter` and then all of the prefixed `liter` variants against `liter` to ensure that they are all correct.

PR #121, opened 2019-04-01
Add more liter-based volume rates.

Aehmlo:
I totally forgot to do this in #116, but it was the main motivation for doing that (d'oh!).

### DIFF
--- a/src/si/volume_rate.rs
+++ b/src/si/volume_rate.rs
@@ -56,6 +56,48 @@ quantity! {
         @cubic_yoctometer_per_second: prefix!(yocto) * prefix!(yocto) * prefix!(yocto);
             "ym³/s", "cubic yoctometer per second", "cubic yoctometers per second";
 
+        @yottaliter_per_second: prefix!(milli) * prefix!(yotta); "YL/s", "yottaliter per second",
+            "yottaliters per second";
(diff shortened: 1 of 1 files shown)


src/si/volume_rate.rs
```diff
@@ -56,6 +56,28 @@ quantity! {
         @cubic_yoctometer_per_second: prefix!(yocto) * prefix!(yocto) * prefix!(yocto);
             "ym³/s", "cubic yoctometer per second", "cubic yoctometers per second";
 
+        @yottaliter_per_second: prefix!(milli) * prefix!(yotta); "YL/s", "yottaliter per second", "yottaliters per second";
```
iliekturtles: Wrap at 100 characters. Just before the plural description.

Aehmlo: Weird that `rustfmt` doesn't catch these.

Aehmlo:
If you're willing, a patch release would be great.

src/si/volume_rate.rs
Aehmlo: Ah, from the [list of limitations](https://github.com/rust-lang/rustfmt#limitations) in the `rustfmt` README:

> Macro declarations and uses (current status: some macro declarations and uses are formatted).

iliekturtles:
Thanks! I'll try to get the changes released later today or tonight.

iliekturtles:
[v0.22.1](https://crates.io/crates/uom) released!

Issue #117, opened 2019-03-24
Adding Torque


dunmatt:
This is a continuation of #114 (which will be closed when #115 goes in) that adds Torque.

I'm starting to look at adding torque, but I have a question.  Over in #114 you mentioned that the new `Kind` will likely want to implement all of the marker traits, but I'm not sure I understand why you said that.  What would it mean to divide a torque by an energy? (for example)

Edit: I think I've figured out the answer to my own question (I'd still appreciate confirmation though).  I think it never derives inter-Kind mathematical operators.  If that's the case, may I suggest changing the docs on the markers?  "Trait to denote that a quantity is able to be X with a quantity of the same dimensions." makes it sound like the Kind isn't taken into account.

iliekturtles:
[Kind](https://jcgm.bipm.org/vim/en/1.2.html) serves two purposes in `uom`. Both to distinguish different quantities of the same dimension that are not mutually comparable (e.g. torque and energy) as well as governing what operations (`+`, `-`, `*`, ...) can be used on the quantity. In torque's case we don't want it comparable with energy but we still want to be able to add torques together as well as perform other operations.

The definition should probably look like this:

```rust
pub trait TorqueKind: // still haven't done any research to see what the appropriate name would be.
    marker::Add + marker::AddAssign
    + marker::Sub + marker::SubAssign
    + marker::Mul + marker::MulAssign
    + marker::Div + marker::DivAssign
    + marker::Rem + marker::RemAssign
    + marker::Neg + marker::Saturating
{
}
```

This all works because [`Kind`](https://github.com/iliekturtles/uom/blob/372f16ed2ea126883b36bf45382fe00fbe0878c3/src/system.rs#L125) is one of the associated types of `Dimension`. `impl` blocks for traits like `Add` or `Eq` only define a single `Dimension` generic parameter ensuring that you can only add or compare quantities with the exact same`Dimension` and therefore exact same `Kind`. Other `impl` blocks such as for `Mul` have a generic parameter for each quantity's `Dimension` but constrain the `impl` to kinds that implement the `marker::Mul` trait.

dunmatt:
That makes sense, and it's consistent with my edit above.  The next question, then, is how are lines like

    let _: Torque<V> = Force::new::<f::newton>(V::one()) * Length::new::<l::meter>(V::one());

supposed to compile?  Right now it's giving me a type mismatch on account of Torque's kind (using the example kind you supplied).

iliekturtles:
[...]

Issue #58, opened 2018-03-23
Implement From trait


iliekturtles:
`impl<D, Ul, Ur, V> From<Quantity<D, Ur, V>> for Quantity<D, Ul, V> ...`

Allow for explicit conversion between quantities with different base units.

iliekturtles:
May require specialization. May require [RFC 1834](https://github.com/rust-lang/rfcs/issues/1834). I don't think there is a way to constrain `typeof(Ul) <> typeof(Ur)` currently. Code available in [dev-from](https://github.com/iliekturtles/uom/tree/dev-from) branch.

```
error[E0119]: conflicting implementations of trait `std::convert::From<si::Quantity<_, _, _>>` for type `si::Quantity<_, _, _>`:
   --> src\system.rs:864:9
    |
864 | /         impl<D, Ul, Ur, V> $crate::lib::convert::From<Quantity<D, Ur, V>> for Quantity<D, Ul, V>
865 | |         where
866 | |             D: Dimension + ?Sized,
867 | |             Ul: Units<V> + ?Sized,
...   |
877 | |             }
878 | |         }
    | |_________^
    |
   ::: src\si\mod.rs
    |
10  | / system! {
11  | |     /// [International System of Quantities](http://jcgm.bipm.org/vim/en/1.6.html) (ISQ).
12  | |     quantities: ISQ {
13  | |         length: meter, L;
...   |
37  | |     }
38  | | }
    | |_- in this macro invocation
    |
    = note: conflicting implementation in crate `core`:
            - impl<T> std::convert::From<T> for T;
```

iliekturtles:
https://www.reddit.com/r/rust/comments/87l21n/conflicting_implementation_when_implementing_from/

iliekturtles:
`impl<U, V> From<Time<U, V>> for Duration` should do it. See https://github.com/iliekturtles/uom/blob/dev-ratio/src/si/ratio.rs#L32 for an example from the `dev-ratio` branch.

dunmatt:
Definitely agree with the above two comments.  Shall I put together a PR?

Aehmlo:
I'd love that; I totally forgot to implement this! I'd be happy to help if you hit any roadblocks, as well.

dunmatt:
I am indeed hitting a roadblock; I don't know a good-enough way to convert a Time into a number of seconds.  `t.get::<second>()` doesn't do it because second somehow doesn't implement the appropriate Conversion (so it doesn't build).

[...]

PR #124, opened 2019-04-20
added torque as an alias of energy

dunmatt:
After a lot of experimentation I'm coming around to the idea that Moments are more trouble than they're worth.  So in that vein, I'm following your suggestion in #117 to not use a different `Kind` for torque, and to add it as a quantity.

To get that to work I had to refactor the quantity macro a bit to split the parts that must be repeated between aliases from the parts that cannot be.  I'm rather new to macros (I still find them very hard to read), so please bear with me if it takes some back and forth to clean these changes up.

### DIFF
--- a/src/quantity.rs
+++ b/src/quantity.rs
@@ -115,6 +115,25 @@ macro_rules! quantity {
     };
     (
         $(#[$quantity_attr:meta])* quantity: $quantity:ident; $description:expr;
+        alias_of: $alias_of:ident;
+        $(#[$dim_attr:meta])* dimension: $system:ident<$($dimension:ident),+>;
+        units {
+            $($(#[$unit_attr:meta])* @$unit:ident: $($conversion:expr),+;
+                $abbreviation:expr, $singular:expr, $plural:expr;)+
+        }
+    ) => {
+        quantity! {
+            $(#[$quantity_attr])* quantity: $quantity; $description;
+            alias_of: $alias_of;
+            $(#[$dim_attr])* dimension: $system<$($dimension),+>;
+            kind: $crate::Kind;
+            units {
+                $($(#[$unit_attr])* @$unit: $($conversion),+; $abbreviation, $singular, $plural;)+
+            }
+        }
+    };
+    (
+        $(#[$quantity_attr:meta])* quantity: $quantity:ident; $description:expr;
         $(#[$dim_attr:meta])* dimension: $system:ident<$($dimension:ident),+>;
         kind: $kind:ty;
         units {
@@ -122,15 +141,14 @@ macro_rules! quantity {
                 $singular:expr, $plural:expr;)+
         }
     ) => {
-        $(#[$dim_attr])*
-        pub type Dimension = super::$system<$($crate::typenum::$dimension),+, $kind>;
-
-        $(#[$quantity_attr])*
-        pub type $quantity<U, V> = super::Quantity<Dimension, U, V>;
-
-        /// Marker trait to identify measurement units for the quantity. See
-        /// [`Unit`](../trait.Unit.html).
-        pub trait Unit: super::Unit {}
+        quantity! { @common
+            $(#[$quantity_attr])* quantity: $quantity; $description;
+            $(#[$dim_attr])* dimension: $system<$($dimension),+>;
+            kind: $kind;
+            units {
+                $($(#[$unit_attr])* @$unit: $($conversion),+; $abbreviation, $singular, $plural;)+
+            }
+        }
 
         /// Trait to identify [units][units] which have a [conversion factor][factor] for the
         /// `Quantity`. See [`Conversion<V>`](../../trait.Conversion.html).
@@ -141,136 +159,6 @@ macro_rules! quantity {
         where
             V: $crate::Conversion<V>,
         {
-        }
-
-        $(quantity!(@unit $(#[$unit_attr])* @$unit);
-
-        impl super::Unit for $unit {
-            #[inline(always)]
-            fn abbreviation() -> &'static str {
-                $abbreviation
-            }
-
-            #[inline(always)]
-            fn singular() -> &'static str {
-                $singular
-            }
-
-            #[inline(always)]
-            fn plural() -> &'static str {
-                $plural
-            }
-        }
-
-        impl Unit for $unit {})+
-
-        storage_types! {
-            types: Float;
-
-            $(impl $crate::Conversion<V> for super::$unit {
-                type T = V;
-
-                #[inline(always)]
-                fn coefficient() -> Self::T {
-                    quantity!(@coefficient $($conversion),+)
-                }
-
-                #[inline(always)]
-                fn constant() -> Self::T {
-                    quantity!(@constant $($conversion),+)
-                }
-            }
-
-            impl super::Conversion<V> for super::$unit {})+
-        }
-
-        storage_types! {
-            types: PrimInt, BigInt;
-            pub type T = $crate::num::rational::Ratio<V>;
-
-            #[inline(always)]
-            fn from_f64(value: f64) -> T {
-                <T as $crate::num::FromPrimitive>::from_f64(value).unwrap()
-            }
-
-            $(impl $crate::Conversion<V> for super::$unit {
-                type T = T;
-
-                #[inline(always)]
-                fn coefficient() -> Self::T {
-                    from_f64(quantity!(@coefficient $($conversion),+))
-                }
-
-                #[inline(always)]
-                fn constant() -> Self::T {
-                    from_f64(quantity!(@constant $($conversion),+))
-                }
-            }
-
-            impl super::Conversion<V> for super::$unit {})+
-        }
-
-        storage_types! {
-            types: BigUint;
-            pub type T = $crate::num::rational::Ratio<V>;
-
-            #[inline(always)]
-            fn from_f64(value: f64) -> T {
-                use $crate::num::FromPrimitive;
-
-                let c = $crate::num::rational::Ratio::<$crate::num::BigInt>::from_f64(value)
-                    .unwrap();
-
-                T::new(c.numer().to_biguint().unwrap(), c.denom().to_biguint().unwrap())
-            }
-
-            $(impl $crate::Conversion<V> for super::$unit {
-                type T = T;
-
-                #[inline(always)]
-                fn coefficient() -> Self::T {
-                    from_f64(quantity!(@coefficient $($conversion),+))
-                }
-
-                #[inline(always)]
-                fn constant() -> Self::T {
-                    from_f64(quantity!(@constant $($conversion),+))
-                }
-            }
-
-            impl super::Conversion<V> for super::$unit {})+
-        }
-
-        storage_types! {
-            types: Ratio;
-
-            #[inline(always)]
-            fn from_f64(value: f64) -> V {
-                <V as $crate::num::FromPrimitive>::from_f64(value).unwrap()
-            }
-
-            $(impl $crate::Conversion<V> for super::$unit {
-                type T = V;
-
-                #[inline(always)]
-                fn coefficient() -> Self::T {
-                    from_f64(quantity!(@coefficient $($conversion),+))
-                }
-
-                #[inline(always)]
-                fn constant() -> Self::T {
-                    from_f64(quantity!(@constant $($conversion),+))
-                }
-            }
-
-            impl super::Conversion<V> for super::$unit {})+
-        }
-
-        /// Quantity description.
-        #[allow(dead_code)]
-        #[inline(always)]
-        pub fn description() -> &'static str {
-            $description
         }
 
         impl<U, V> $quantity<U, V>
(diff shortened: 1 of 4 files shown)


src/quantity.rs
```diff
@@ -113,6 +113,25 @@ macro_rules! quantity {
             }
         }
     };
+    (
+        $(#[$quantity_attr:meta])* quantity: $quantity:ident; $description:expr;
+        alias_of: $alias_of:ident;
```
iliekturtles: Why add the alias? In my first brief look at the code it looks like it allows the alias's units to be used?

Quantities with the same dimension and kind are already type aliases although I'm they can't share units currently.

```rust
type Energy = Quantity<m^2 * kg * s^-2>;
type Torque = Quantity<m^2 * kg * s^-2>;
```

src/si/mod.rs
```diff
@@ -49,6 +49,7 @@ system! {
         temperature_interval::TemperatureInterval,
         thermodynamic_temperature::ThermodynamicTemperature,
         time::Time,
+        torque_magnitude::TorqueMagnitude,
```
iliekturtles: I'm inclined towards `Torque` instead of `TorqueMagnitude`. See `Force` as an existing example. Additionally I'd like to get support for mathematical vectors as the underlying storage type some day (e.g. `Quantity<V = Vec3f32>` rather than `Vec3f32<T = Quantity>`). Thoughts?

src/si/torque_magnitude.rs
```diff
@@ -0,0 +1,155 @@
+//! Torque (aka moment of force) (base unit newton meter, kg<sup>1</sup> · m<sup>2</sup> ·
+//! s<sup>-2</sup>).
+
+quantity! {
+    /// Torque magnitude (base unit newton meter, kg<sup>1</sup> · m<sup>2</sup> · s<sup>-2</sup>).
+    ///
+    /// Torques are moments, by which I mean they inherently depend on a distance to a frame of
```
iliekturtles: Use implicit second person for comments (https://developers.google.com/style/person). I think that is the correct description. Don't use "I".

src/si/torque_magnitude.rs
```diff
@@ -0,0 +1,155 @@
...
+        @newton_centimeter: prefix!(centi); "N · cm", "newton centimeter", "newton centimeters";
+        @newton_millimeter: prefix!(milli); "N · mm", "newton millimeter", "newton millimeters";
+        @newton_micrometer: prefix!(micro); "N · µm", "newton micrometer", "newton micrometers";
+        @newton_nanometer: prefix!(nano); "N · nm", "newton nanometer", "newton nanometers";
+        @newton_picometer: prefix!(pico); "N · pm", "newton picometer", "newton picometers";
+        @newton_femtometer: prefix!(femto); "N · fm", "newton femtometer", "newton femtometers";
+        @newton_attometer: prefix!(atto); "N · am", "newton attometer", "newton attometers";
+        @newton_zeptometer: prefix!(zepto); "N · zm", "newton zeptometer", "newton zeptometers";
+        @newton_yoctometer: prefix!(yocto); "N · ym", "newton yoctometer", "newton yoctometers";
+
```
iliekturtles: Just one blank line between sections.

src/si/torque_magnitude.rs
```diff
@@ -0,0 +1,155 @@
+//! Torque (aka moment of force) (base unit newton meter, kg<sup>1</sup> · m<sup>2</sup> ·
```
iliekturtles: List base quantities in the same order as specified in the `system!` macro: length, mass, time, .... Superscript isn't necessary when equal to `1`.

src/quantity.rs
dunmatt: It's so the `Conversion` can extend the `Unit` of both the original type and the alias.

src/si/torque_magnitude.rs
dunmatt: Shall I update Force as well?  (that's where I copy pasted from)

src/si/mod.rs
dunmatt: I've done the renaming.  In terms of planning for vectors, I'll spend some time thinking about it and let you know what would be a best case scenario for my uses.

src/si/torque_magnitude.rs
iliekturtles: Added #127 regarding units. Reply there / submit a new PR if you're interested. Otherwise I'll do in the next few days. We'll also want to make kg superscript consistent with that issue. Leaning towards leaving the 1 right now.

iliekturtles:
Changes in `torque.rs` look pretty good. Still needs a full detailed review, but I don't expect that to take long. Also need to fully understand the `alias_of` changes. I'm hoping to get to all of this over the weekend.

src/si/mod.rs
dunmatt: So, I spent some time poking around and it seems like nobody has written the vector library I want, yet.  It's very tempting to launch yet another hobby project.  What I really want is compile time checking of transform matrix math, to prevent nonsense matrix combinations (ie if Tab transforms a point or vector from a to b, I want `Tab*Tbc == Tac` and both `Tab*Tcb` and `Tbc*Tab` should break the build)... but I don't strictly need them to be quantities, all I really care about is compile time frame of reference checking.

I'm curious though, what's your motivation for wanting a vector quantity (as opposed to a vectory of quantities)?

iliekturtles: I figured a `Quantity<V = Vec3f32>` would be more natural to use than `Vec3<T = Quantity>`. No actual usage experience and I'm not looking to attempt to implement now unless someone actually has a use case.

iliekturtles:
Finally got to look at this with the compiler and I understand what is happening now. `Energy` and `Torque` are both aliases for `Quantity<Dimension<P2, P1, N2>, U, V>`. The `quantity!` macro implements `FromStr` which causes a conflicting implementations error shown below.

A short-term solution could be to revert the `alias_of` changes in quantity.rs and give `Torque` a kind so that it becomes distinct from `Energy`. A `From<Energy> for Torque` implementation would likely also be needed. I don't think this is the right long-term solution but we could do it right now and include `Torque` in the next `uom` release. Once a better long-term solution is found the kind could be dropped.

Another possible solution would be to change the `FromStr` implementation to return a type parameterized on the quantity's `Unit` trait so that parsing would be something like `let t: Torque = "1 N".parse::<(Torque, torque::Unit)>().into()`. I don't think this exact form would work so more experimentation is needed.

```bash
error[E0119]: conflicting implementations of trait `std::str::FromStr` for type `si::Quantity<(dyn si::D
imension<M=typenum::PInt<typenum::UInt<typenum::UTerm, typenum::B1>>, J=typenum::Z0, L=typenum::PInt<typ
enum::UInt<typenum::UInt<typenum::UTerm, typenum::B1>, typenum::B0>>, T=typenum::NInt<typenum::UInt<type
num::UInt<typenum::UTerm, typenum::B1>, typenum::B0>>, Th=typenum::Z0, N=typenum::Z0, Kind=(dyn Kind + '
static), I=typenum::Z0> + 'static), _, f32>`:
   --> src\quantity.rs:348:17
    |
348 |                   impl<U> FromStr for super::super::$quantity<U, V>
    |  _________________^
    | |_________________|
    | |
349 | |                 where
350 | |                     U: super::super::super::Units<V> + ?Sized,
351 | |                 {
...   |
364 | |                     }
365 | |                 }
    | |                 ^
    | |_________________|
    | |_________________first implementation here
    |                   conflicting implementation for `si::Quantity<(dyn si::Dimension<M=typenum::PInt<
typenum::UInt<typenum::UTerm, typenum::B1>>, J=typenum::Z0, L=typenum::PInt<typenum::UInt<typenum::UInt<
typenum::UTerm, typenum::B1>, typenum::B0>>, T=typenum::NInt<typenum::UInt<typenum::UInt<typenum::UTerm,
 typenum::B1>, typenum::B0>>, Th=typenum::Z0, N=typenum::Z0, Kind=(dyn Kind + 'static), I=typenum::Z0> +
 'static), _, f32>`
    |
   ::: src\si\torque.rs:3:1
    |
3   | / quantity! {
4   | |     /// Torque magnitude (base unit newton meter, kg<sup>1</sup> · m<sup>2</sup> · s<sup>-2</sup
>).
5   | |     ///
6   | |     /// Torques are moments, which means they inherently depend on a distance to a frame of
...   |
75  | |     }
76  | | }
```

dunmatt:
At least for my edification, what's the problem with the way it's implemented now?

iliekturtles:
With the current `alias_of` code `Torque::from_str` will only parse `Energy` abbreviations and use those conversion factors. Some other methods like `new`, `get`, `floor, ... look like they are only implemented for the parent and not the alias, although I didn't test.

We need to find some way to separate `Energy` and `Torque` `from_str` implementations so they can each use their own unit definitions. Finding some way for the two to share units or be sub/super-sets of each other is another option and then share a `from_str` implementation.

dunmatt:
How would you feel about a blanked From implementation that can switch between `Kind`s?

iliekturtles:
A blanket implementation probably makes sense. I've been thinking about this a little bit today and I'm not sure if it makes sense to ever have different quantities, even if they are the same dimension to be strict aliases of themselves.

```rust
let e: Energy = ...;

energy_fn(e);
torque_fn(e);

fn energy_fn(e: Energy) {}
fn torque_fn(t: Torque) {}
```

The above works without any warnings and could easily lead to mixing up quantities of the same dimension. A blanket implementation would at least require a `.into()` telling the user a type change was happening.

My thoughts aren't fully formed yet, but likely more design effort needs to be put into what is called kind right now.

dunmatt:
I'm working at the edge of my skill here, can you sanity check this?  (so far I haven't been able to get From to allow K1 * K2.into() to work)


        impl<K, U, V> From<super::Quantity<super::$system<$($crate::typenum::$dimension),+, K>, U, V>> for super::Quantity<Dimension, U, V>
        where
            U: super::Units<V>,
            V: num_traits::Num + $crate::Conversion<V>,
        {
            fn from(val: super::Quantity<super::$system<$($crate::typenum::$dimension),+, K>, U, V>) -> super::Quantity<Dimension, U, V> {
                unimplemented!()
            }
        }

iliekturtles:
Sat down to look at this and realized a blanket impl isn't possible. See #58. However an explicit `into()` method can be implemented on `Quantity`. I tested the following in the `impl Quantity<D, U, V>` in `system.rs`. Name conflicts with `Into::into` in other tests and I didn't take time to figure the workaround yet so it's called `iinto` right now. Uncommenting the second `into()` call causes the test to fail.

```rust
            #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
            /// # use uom::si::f32::*;
            /// # use uom::si::thermodynamic_temperature::kelvin;
            /// let tt: ThermodynamicTemperature = ThermodynamicTemperature::new::<kelvin>(1.0);
            /// let ti: TemperatureInterval = tt.iinto();
            /// //let f: Frequency = ti.into();
            /// ```
            #[inline(always)]
            pub fn iinto<Dr, Ur>(self) -> Quantity<Dr, Ur, V>
            where
                $(D::$symbol: $crate::lib::ops::IsEqual<Dr::$symbol>,)+
                Dr: Dimension + ?Sized,
                Ur: Units<V> + ?Sized,
            {
                Quantity {
                    value: change_base::<Dr, U, Ur, V>(&self.value),
                }
            }
```

dunmatt:
I don't understand what #58 has to do with this change.  My From impl above compiles just fine, the problems only arise at the call site because the compiler doesn't know what size anything is.  Can you help me understand the core of the issue?

Is there a way I could solve the specific problem to allow the Torque tests to work so that I can get those in (and start using them) and worry about the more general case later?

iliekturtles:
Regarding #58 I was thinking about a blanket impl in the `system!` macro that would apply to all quantities. This isn't possible because of the overlapping `impl<T> From<T> for T` in the standard library. Doing the impl in the `quantity!` macro like you are doing may get around this. If you add a `K: ?Sized` constraint to your impl can you compile or do you get the same conflicting implementations error?

You should also add two base units parameters (`Ul`, `Ur`) so that `From` can be used to change kinds or base units.

I just skimmed the code so far but will plan to look more in depth tomorrow evening.

dunmatt:
Ok, so adding `K: ?Sized,` got me to the point where I'm now having the conflicting implementations problem you're talking about.  As a potential solution, what if I made a new pattern for the quantity macro that did the standard thing, and then also created From impls that convert things between Kind and AngleKind?  So since the Kinds would be fully specified (and different) it wouldn't conflict with the blanket implementation in std.

As to your Ul and Ur idea.  I don't think it's a good idea to allow converting both of those things as once, if I'm using `into` to create an AngleKind of a Length, I don't want to be able to screw up and accidentally turn an Area into an AngleKind of Length.  If there was a way to have an implementation do one or the other but not both that might be okay, but I don't know how that might be achieved.

iliekturtles:
I was afraid you would get the same conflicting implementation error. Explicitly specifying `Kind`s is probably going to be needed. Let us know how that works.

For changing base units `Ul` to `Ur` this doesn't affect dimension, only the base units a quantity is stored in. e.g. `Length<U<meter>>` to `Length<U<foot>>` (not actual syntax). Go ahead and implement with a single base unit, `U`, first and then we can see if it makes sense to support changing.

Issue #127, opened 2019-04-24
Make quantity dimension documentation consistent


iliekturtles:
Ensure comments use base units are listed in the same order as the base quantities are specified in the `system!` macro. Should units with a superscript of 1 be shown? Looks like most locations do show the superscript but a few don't. I'm leaning towards showing but welcome input.

Please reply here if you're interested in reviewing and making necessary changes. If no one else has noted their interested I plan on doing the review in the next couple days.

No superscripts shown. Should they be?
```rust
//! Electric charge (base unit coulomb, A · s).
```

Superscrit shown. Should it be?
```rust
//! Electric current (base unit ampere, A<sup>1</sup>).
```

Units are out of order. Should be m^1 * kg^1 * s^-2.
```rust
//! Force (base unit newton, kg<sup>1</sup> · m<sup>1</sup> · s<sup>-2</sup>).
```

dunmatt:
Why is the important order for the documentation the one in the macro invocation?  That seems like a leaky abstraction.  What you really care about when writing documentation is what will be the easiest for the readers to understand.  In that light, I would suggest `Force` is not wrong.

The most familiar aspect of force to basically everyone is "F=ma", so it's naturally presented with an amount of mass times an amount of acceleration, kg * (m/s^2), and indeed this is how it is spelled out in Wikipedia.  m/s^2 * kg is worse because it is less familiar, but it's still much better than m * kg/s^2.

I guess what I'm saying is if I had my druthers the documentation would be optimized for the readers, even at the expense of the writers.  As far as the 1 superscripts, I'd say remove them because they're not typically used (even https://www.nist.gov/pml/nist-guide-si-appendix-b9-factors-units-listed-kind-quantity-or-field-science doesn't use them).

Aehmlo:
@dunmatt This isn't so much a "leaky abstraction" as a strict interpretation of the specification. The [SI brochure](https://www.bipm.org/en/publications/si-brochure/section1-3.html) describes the 7 base dimensions exactly as they are arranged in this crate. As such, the *dimensions* in this crate arguably *must* be in the order they're written in.

The *units* in the crate and the ordering thereof is somewhat less black-and-white. From what I've read of the spec, there's not a clear statement of the expected order of units in terms of other units, but a strict interpretation would argue for (and convention seems to be) writing units in the same order as dimensions. I definitely see where you're coming from for stuff such as forces, but I also think that there's great value in consistency.

[...]

PR #129, opened 2019-04-29
Homogenize documentation

Aehmlo:
As discussed in #127. I also went ahead and renamed density to "mass density," as this is more appropriate. If desired, I can separate that change into another PR and maybe set up an alias or something.

There are a couple of places where "dimension of `X`" is weird (especially `Ratio` and `Angle`); do we want to special-case these or just let them sound weird?

### DIFF
--- a/src/si/angle.rs
+++ b/src/si/angle.rs
@@ -1,13 +1,13 @@
-//! Angle (dimensionless).
+//! Angle (dimensionless quantity).
 
 /// AngleKind is a `Kind` for separating angular quantities from their indentically dimensioned
 /// non-angular quantity counterparts.
 pub trait AngleKind: ::Kind {}
 
 quantity! {
-    /// Angle (dimensionless).
+    /// Angle (dimensionless quantity).
     quantity: Angle; "angle";
-    /// Angle dimension (dimensionless).
+    /// Dimension of angle, 1 (dimensionless).
     dimension: ISQ<
         Z0,     // length
         Z0,     // mass
--- a/src/si/acceleration.rs
+++ b/src/si/acceleration.rs
@@ -1,9 +1,9 @@
-//! Acceleration (base unit meter per second<sup>2</sup>, m<sup>1</sup> · s<sup>-2</sup>).
+//! Acceleration (base unit meter per second squared, m · s⁻²).
 
 quantity! {
-    /// Acceleration (base unit meter per second<sup>2</sup>, m<sup>1</sup> · s<sup>-2</sup>).
+    /// Acceleration (base unit meter per second squared, m · s⁻²).
     quantity: Acceleration; "acceleration";
-    /// Acceleration dimension, m<sup>1</sup> · s<sup>-2</sup>.
+    /// Dimension of acceleration, LT⁻² (base unit meter per second squared, m · s⁻²).
     dimension: ISQ<
         P1,     // length
         Z0,     // mass
--- a/src/si/amount_of_substance.rs
+++ b/src/si/amount_of_substance.rs
@@ -1,9 +1,9 @@
-//! Amount of substance (base unit mole, mol<sup>1</sup>).
+//! Amount of substance (base unit mole, mol).
 
 quantity! {
-    /// Amount of substance (base unit mole, mol<sup>1</sup>).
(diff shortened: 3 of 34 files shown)


Aehmlo:
Realized the "mass density" change wasn't completely done and belonged in a separate PR anyway.

src/si/angle.rs
```diff
@@ -1,4 +1,4 @@
-//! Angle (dimensionless).
+//! Angle (unitless).
```
iliekturtles: Unitless isn't the appropriate description. https://jcgm.bipm.org/vim/en/1.8.html Let's use "quantity of dimension one" or "dimensionless quantity."

src/si/angle.rs
```diff
@@ -7,7 +7,7 @@ pub trait AngleKind: ::Kind {}
 quantity! {
     /// Angle (dimensionless).
     quantity: Angle; "angle";
-    /// Angle dimension (dimensionless).
+    /// Dimension of angle (dimensionless, unitless).
```
iliekturtles: I think I'm OK with keeping "Dimension of angle." Will think on it more.

iliekturtles:
I appreciate the effort put into this PR. Apply the patch below and I'll merge! I based my changes on NIST 811 [section 9.6](https://www.nist.gov/pml/nist-guide-si-chapter-9-rules-and-style-conventions-spelling-unit-names) which states "modifiers ... are used and are placed after the unit name. ... the modifiers may, however, be placed before the unit name in the case of area or volume." The way I read it is that the modifier may only appear before the unit in the case of length (area, volume).

```diff
diff --git a/src/si/acceleration.rs b/src/si/acceleration.rs
index c3c8c22..356dba3 100644
--- a/src/si/acceleration.rs
+++ b/src/si/acceleration.rs
@@ -1,9 +1,9 @@
-//! Acceleration (base unit meter per square second, m · s⁻²).
+//! Acceleration (base unit meter per second squared, m · s⁻²).

 quantity! {
-    /// Acceleration (base unit meter per square second, m · s⁻²).
+    /// Acceleration (base unit meter per second squared, m · s⁻²).
     quantity: Acceleration; "acceleration";
-    /// Dimension of acceleration, LT⁻² (base unit meter per square second, m · s⁻²).
+    /// Dimension of acceleration, LT⁻² (base unit meter per second squared, m · s⁻²).
     dimension: ISQ<
         P1,     // length
         Z0,     // mass
/s/uom (Aehmlo-docs *$)$ git diff
diff --git a/src/si/acceleration.rs b/src/si/acceleration.rs
index c3c8c22..356dba3 100644
--- a/src/si/acceleration.rs
+++ b/src/si/acceleration.rs
@@ -1,9 +1,9 @@
-//! Acceleration (base unit meter per square second, m · s⁻²).
+//! Acceleration (base unit meter per second squared, m · s⁻²).

 quantity! {
-    /// Acceleration (base unit meter per square second, m · s⁻²).
+    /// Acceleration (base unit meter per second squared, m · s⁻²).
     quantity: Acceleration; "acceleration";
-    /// Dimension of acceleration, LT⁻² (base unit meter per square second, m · s⁻²).
+    /// Dimension of acceleration, LT⁻² (base unit meter per second squared, m · s⁻²).
     dimension: ISQ<
         P1,     // length
         Z0,     // mass
diff --git a/src/si/jerk.rs b/src/si/jerk.rs
index a485958..17c8366 100644
--- a/src/si/jerk.rs
+++ b/src/si/jerk.rs
@@ -1,9 +1,9 @@
-//! Jerk (base unit meter per cubic second, m · s⁻³).
+//! Jerk (base unit meter per second cubed, m · s⁻³).

 quantity! {
-    /// Jerk (base unit meter per cubic second, m · s⁻³).
+    /// Jerk (base unit meter per second cubed, m · s⁻³).
     quantity: Jerk; "jerk";
-    /// Dimension of jerk, LT⁻³ (base unit meter per cubic second, m · s⁻³).
+    /// Dimension of jerk, LT⁻³ (base unit meter per second cubed, m · s⁻³).
     dimension: ISQ<
         P1,     // length
         Z0,     // mass
```

iliekturtles:
Thanks!

PR #130, opened 2019-04-29
Rename density to mass density

Aehmlo:
"Mass density" is the more rigorous and technically correct version (though I believe the SI brochure does essentially say they're equivalent), so I think we should at least expose this as an option.

To retain backwards compatibility, this PR contains `pub type Density<U, V> = MassDensity<U, V>`. I'd rather not have this from an idealistic standpoint, but from a practical perspective, it makes sense to not break backwards compatibility for this (which I would expect to be one of the more commonly-used quantities). We could also just make the type alias work the other way around (`pub type MassDensity<U, V> = Density<U, V>`) and thereby obtain a single-line diff (two lines with docs), but I do think that this is the more semantically correct version.


iliekturtles:
I agree that `MassDensity` is the better name. I can't remember why `Density` was chosen over `MassDensity` but I'll do a bit of digging when I get a moment. I think it would also be good to put a [`#[deprecated]`](https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md) attribute on the alias to the old name in order to get users to switch.

Aehmlo:
Good idea, I’ll add that.

src/si/density.rs
```diff
@@ -1,9 +1,13 @@
-//! Density (base unit kilogram per cubic meter, m<sup>-3</sup> · kg<sup>1</sup>).
+//! Mass density (base unit kilogram per cubic meter, m<sup>-3</sup> · kg<sup>1</sup>).
+
+/// Mass density (base unit kilogram per cubic meter, m<sup>-3</sup> · kg<sup>1</sup>).
+#[deprecated(since = "0.22.3", note = "Please use the more descriptive MassDensity instead")]
```
iliekturtles: Period at the end of note.

src/si/mod.rs
```diff
@@ -27,7 +27,7 @@ system! {
         area::Area,
         available_energy::AvailableEnergy,
         capacitance::Capacitance,
-        density::Density,
+        density::MassDensity,
```
iliekturtles: Rename module / filename to `mass_density.rs`.

iliekturtles:
Thanks! I manually merged and resolved the conflicts with #129.

PR #131, opened 2019-05-01
Added an implementation of From<Time> for Duration

dunmatt:
This resolves a request in https://github.com/iliekturtles/uom/issues/58#issuecomment-383260348

### DIFF
--- a/src/si/time.rs
+++ b/src/si/time.rs
@@ -49,12 +49,31 @@ quantity! {
 
 impl<U, V> From<Time<U, V>> for Duration
 where
-    U: ::si::Units<V>,
+    U: ::si::Units<V> + ?Sized,
     V: ::num::Num + ::num::AsPrimitive<f64> + ::Conversion<V>,
 {
     fn from(t: Time<U, V>) -> Duration {
         let secs: f64 = t.value.as_();
         let nanos = (secs * 1e9) as u64 % 1e9 as u64;
         Duration::new(secs as u64, nanos as u32)
+    }
+}
+
+#[cfg(all(test, feature = "std"))]
+mod tests {
+    storage_types! {
+        types: Float;
+
+        use ::lib::time::Duration;
+        use si::quantities::*;
+        use si::time::second;
+
+        #[test]
+        fn from() {
+            let t = Time::new::<second>(21.5);
+            let d: Duration = t.into();
+            assert_eq!(21, d.as_secs());
+            assert_eq!(5e8 as u32, d.subsec_nanos());
+        }
     }
 }
--- a/src/si/time.rs
+++ b/src/si/time.rs
@@ -54,12 +54,10 @@ where
     second: ::Conversion<V, T = V::T>,
     nanosecond: ::Conversion<V, T = V::T>,
 {
-    fn from(t: Time<U, V>) -> Duration {
-        let t = if t < Time::<U, V>::new::<second>(V::zero()) {
-            Time::<U, V>::new::<second>(V::zero()) - t
-        } else {
-            t
-        };
+    fn from(mut t: Time<U, V>) -> Duration {
+        if t < Time::<U, V>::new::<second>(V::zero()) {
+            t = Time::<U, V>::new::<second>(V::zero()) - t
+        }
         let secs = t.get::<second>().as_();
         let nanos = (t % Time::<U, V>::new::<second>(V::one())).get::<nanosecond>().as_();
         Duration::new(secs, nanos)
@@ -85,11 +83,6 @@ mod tests {
             let t = Time::new::<second>(12.345);
             let d: Duration = t.into();
             assert_eq!(d.as_secs(), 12);
-            assert!(344_999_500 <= d.subsec_nanos() && d.subsec_nanos() <= 345_000_500);
-
-            let t = Time::new::<second>(0.345);
-            let d: Duration = t.into();
-            assert_eq!(d.as_secs(), 0);
             assert!(344_999_500 <= d.subsec_nanos() && d.subsec_nanos() <= 345_000_500);
         }
     }
--- a/src/si/time.rs
+++ b/src/si/time.rs
@@ -81,6 +81,12 @@ mod tests {
             let d: Duration = t.into();
             assert_eq!(d.as_secs(), 21);
             assert!(499_999_999 <= d.subsec_nanos() && d.subsec_nanos() <= 500_000_001);
+
+            let t = Time::new::<second>(12.345);
+            let d: Duration = t.into();
+            assert_eq!(d.as_secs(), 12);
+            assert_eq!(d.subsec_nanos(), 345_000_000);
+            // assert!(499_999_999 <= d.subsec_nanos() && d.subsec_nanos() <= 500_000_001);
         }
     }
 }
--- a/src/si/time.rs
+++ b/src/si/time.rs
@@ -53,7 +53,7 @@ where
     V: ::num::Num + ::num::AsPrimitive<f64> + ::Conversion<V>,
 {
     fn from(t: Time<U, V>) -> Duration {
-        let secs: f64 = t.value.as_();
+        let secs = t.value.as_().abs();
         let nanos = (secs * 1e9) as u64 % 1e9 as u64;
         Duration::new(secs as u64, nanos as u32)
     }
@@ -70,7 +70,7 @@ mod tests {
 
         #[test]
         fn from() {
-            let t = Time::new::<second>(21.5);
+            let t = Time::new::<second>(-21.5);
             let d: Duration = t.into();
             assert_eq!(21, d.as_secs());
             assert_eq!(5e8 as u32, d.subsec_nanos());
--- a/src/si/time.rs
+++ b/src/si/time.rs
@@ -85,8 +85,7 @@ mod tests {
             let t = Time::new::<second>(12.345);
             let d: Duration = t.into();
             assert_eq!(d.as_secs(), 12);
-            assert_eq!(d.subsec_nanos(), 345_000_000);
-            // assert!(499_999_999 <= d.subsec_nanos() && d.subsec_nanos() <= 500_000_001);
+            assert!(344_999_500 <= d.subsec_nanos() && d.subsec_nanos() <= 345_000_500);
         }
     }
 }
--- a/src/si/time.rs
+++ b/src/si/time.rs
@@ -50,12 +50,18 @@ quantity! {
 impl<U, V> From<Time<U, V>> for Duration
 where
     U: ::si::Units<V> + ?Sized,
-    V: ::num::Num + ::num::AsPrimitive<f64> + ::Conversion<V>,
+    V: ::num::Num + ::num::AsPrimitive<u64> + ::num::AsPrimitive<u32> + ::Conversion<V> + ::lib::cmp::PartialOrd,
+    second: ::Conversion<V, T = V::T>,
+    nanosecond: ::Conversion<V, T = V::T>,
 {
     fn from(t: Time<U, V>) -> Duration {
-        let secs = t.value.as_().abs();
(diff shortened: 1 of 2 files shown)


iliekturtles:
Thanks for the PR! I'm planning to review tonight along with responding to the other open conversations.

dunmatt:
Trying to write a test for this I'm running into the same ?Sized problem as in the Torque PR.  Is there a problem with my test?


        #[cfg(all(test, feature = "std"))]
        mod tests {
            storage_types! {
                types: Float;

                use ::lib::time::Duration;
                use si::quantities::*;
                use si::time::second;
                use tests::Test;

                quickcheck! {
                    // #[allow(trivial_casts)]
                    fn from(v: V) -> bool {
                        let t = Time::new::<second>(v);
                        let d: Duration = t.into();
                        Test::eq(v as u64, d.secs)
                        // Test::eq(&l.hypot(r),
                        //     &Length::new::<meter>(l).hypot(Length::new::<meter>(r)).get::<meter>())
                    }
                }
            }
        }

iliekturtles:
`Units<V>` also needs the `?Sized` constraint: https://github.com/iliekturtles/uom/blob/8c5eb43c4de5870ff3607ecf8ee2fd49b994d854/src/quantity.rs#L278

src/si/time.rs
```diff
@@ -44,3 +46,34 @@ quantity! {
...
 }
+
+impl<U, V> From<Time<U, V>> for Duration
+where
+    U: ::si::Units<V> + ?Sized,
+    V: ::num::Num + ::num::AsPrimitive<f64> + ::Conversion<V>,
+{
+    fn from(t: Time<U, V>) -> Duration {
+        let secs = t.value.as_().abs();
+        let nanos = (secs * 1e9) as u64 % 1e9 as u64;
```
iliekturtles: I'm worried about overflow here. It would be good to find a way to extract the nanoseconds without multiplication. Does converting the original value into nanoseconds and then extracting the portion needed for `Duration` work?

src/si/time.rs
```diff
@@ -44,3 +46,34 @@ quantity! {
...
     }
 }
+
+impl<U, V> From<Time<U, V>> for Duration
+where
+    U: ::si::Units<V> + ?Sized,
+    V: ::num::Num + ::num::AsPrimitive<f64> + ::Conversion<V>,
+{
+    fn from(t: Time<U, V>) -> Duration {
+        let secs = t.value.as_().abs();
```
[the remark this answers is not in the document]
dunmatt: I also tried the obvious thing:

    fn from(t: Time<U, V>) -> Duration {
        let secs = t.get::<second>();
        let nanos = t.get::<nanosecond>();
        Duration::new(secs as u64, nanos as u32)
    }

but I haven't yet found a way to get it to compile.  TBH I'm pretty frustrated with how hard even basic things are.  I try to do a thing, and then I invariably do it wrong because I can't make sense of the current code.  I tried calling from_base, but there were enough unsatisfied trait bound errors that my brain kinda switched off.

It might just be that I'm not good enough to contribute here, because this conversion ought to be the simplest thing in the world to implement.

src/si/time.rs
dunmatt: Converting the original value to nanoseconds is a multiplication operation... there's no way around that part.  That said, taking off the integer part before multiplying can solve any potential overflow problems:  `let nanos = secs.fract() * 1e9;`.  It doesn't solve any of the base problems in your other point, however...

iliekturtles:
It's been almost three years since I started the project and I still spend an excessive amount of time staring at compiler errors trying to figure things out so you're definitely not alone.

I got the code compiling, but it doesn't pass tests. Now to find a correct, and hopefully efficient, algorithm to convert into a `Duration`!

 * I was hoping to delay the conversion to `u64`/`u32` to as late as possible and not go through an intermediate type (`f64`) if possible. This may require multiple implementations via `storage_types!` in order to satisfy the fractional/integer and signed/unsigned differences. It may also be a better use of time to land the simplest algorithm now.
 * I added a modulo prior to the nanosecond conversion. Multiplication avoided! Not sure what other issues this introduces. Precision issues?
 * Test doesn't seem to require std so I removed.
 * It would be good to see if this could be made into a `quickcheck` test. Maybe use the `Duration::from_nanos` constructor and compare against `Time::<nanosecond>::new(v).into()`?
 * Testing with different base units would be good but I'm not sure what kind of compile time hit it would be calling the `ISQ!` macro a couple times.

```diff
diff --git a/src/si/time.rs b/src/si/time.rs
index 4f267a2..b6fbd8f 100644
--- a/src/si/time.rs
+++ b/src/si/time.rs
@@ -50,16 +50,19 @@ quantity! {
 impl<U, V> From<Time<U, V>> for Duration
 where
     U: ::si::Units<V> + ?Sized,
-    V: ::num::Num + ::num::AsPrimitive<f64> + ::Conversion<V>,
+    V: ::num::Num + ::num::AsPrimitive<u64> + ::num::AsPrimitive<u32> + ::Conversion<V>,
+    second: ::Conversion<V, T = V::T>,
+    nanosecond: ::Conversion<V, T = V::T>,
 {
     fn from(t: Time<U, V>) -> Duration {
-        let secs = t.value.as_().abs();
-        let nanos = (secs * 1e9) as u64 % 1e9 as u64;
-        Duration::new(secs as u64, nanos as u32)
+        let secs = t.get::<second>().as_();
+        let nanos = (t % Time::<U, V>::new::<second>(V::one())).get::<nanosecond>().as_();
+
+        Duration::new(secs, nanos)
     }
 }

-#[cfg(all(test, feature = "std"))]
+#[cfg(test)]
 mod tests {
     storage_types! {
         types: Float;
@@ -72,8 +75,9 @@ mod tests {
         fn from() {
             let t = Time::new::<second>(-21.5);
             let d: Duration = t.into();
+
             assert_eq!(21, d.as_secs());
-            assert_eq!(5e8 as u32, d.subsec_nanos());
+            assert_eq!(500_000_000, d.subsec_nanos());
         }
     }
 }
```

dunmatt:
Resuming work, the reason your test failed was because you didn't call `t.abs()`.  The trouble I'm working on now is that `abs` is a method on `num_traits::Signed`, and for whatever reason `num_traits::Signed` and `num_traits::Unsigned` are not disjoint, so you get conflicting impls if you naively try to only call `abs` when it's present.

Your "use quickcheck" idea is why the tests module was gated on std; I was copying from Length (but took out quickcheck during the compiler struggles).

dunmatt:
Ok, so I thought challenges couldn't get easier that implementing this conversion... then I attempted to implement `abs` and failed.  Check out my most recent commit, there is no way this isn't a bug somewhere else in the code.  I put in -21 and it gives me back 24!

src/si/time.rs
```diff
@@ -55,17 +55,18 @@ where
     nanosecond: ::Conversion<V, T = V::T>,
 {
     fn from(t: Time<U, V>) -> Duration {
-        let mut secs = t.get::<second>();
-        if secs < V::zero() {
-            secs = V::zero() - secs;
-        }
-        let secs = secs.as_();
+        let t = if t < Time::<U, V>::new::<second>(V::zero()) {
```
Aehmlo: I understand why this is here, but do we want to silently convert negative times to (inherently positive) durations? I suppose this might be a matter of "time vs. time interval."

src/si/time.rs
```diff
@@ -76,11 +77,10 @@ mod tests {
 
         #[test]
         fn from() {
-            // let t = Time::new::<second>(21.5);
             let t = Time::new::<second>(-21.5);
             let d: Duration = t.into();
-            assert_eq!(21, d.as_secs());
-            // assert_eq!(5e8 as u32, d.subsec_nanos());
+            assert_eq!(d.as_secs(), 21);
+            assert!(499_999_999 <= d.subsec_nanos() && d.subsec_nanos() <= 500_000_001);
```
Aehmlo: Is there a reason that we can't require this to be exactly 500,000,000?

dunmatt: Precision loss when I subtract t from zero is my current best guess.  The way I implemented it originally, by casting the time to f64 and then converting that to a Duration did allow this to be 500,000,000 exactly.  Perhaps it's only the f32 test that loses precision?

src/si/time.rs
dunmatt: I don't think it's really a silent conversion, the user has to manually call `into` in a context where the conversion is into an inherently positive type.  I guess it's a question of which is least surprising:

1) Absolute Value
2) Wrapping Around
3) Minimum 0

Aehmlo: I know that I would prefer panicking, although I’d imagine that others would see it differently.

dunmatt: Really?  I just put that option there for completeness, From [**must not fail**](https://doc.rust-lang.org/std/convert/trait.From.html)

Aehmlo: In the context of that documentation page, "fallible functions" seem to be loosely be "functions that return `Result<T, E>` or `Option<T>`:

> When constructing a function that is capable of failing the return type will generally be of the form `Result<T, E>`.

As such, I read that note as simply meaning "`From` does not return an error type."  A more general argument for this interpretation is that it's almost nonsensical to say "this code must not panic" (with a couple of notable exceptions) in Safe Rust.

Panics signal an irrecoverable error, something that makes no sense and for which there's no sensible alternative route. In my mind, converting a negative time to a `Duration` is such an error, and it makes sense to panic, as this likely signals a logic error. However, panicking from within a library crate is certainly not ideal.

I actually might argue, if possible, that we should use `TryFrom` instead of `From`.

dunmatt: So `TryFrom` vs `From` is a different issue, there's certainly an argument to be made there.  The trick is, doing that sort of error management (while certainly more rigorous) is enough of an ergonomics hit that if that's the direction `uom` decides to go I'm going to just continue defining my own conversion functions where and as I need them.  I'm not saying you're wrong, you're not, `TryFrom` is definitely a more correct fit, it just feels cumbersome (and the implementation is going to be even more of a nightmare).

So with that out of the way, I think your reading of the note is wrong, and here's why.  There's absolutely no point saying "`From` does not return an error type", because there is no flexibility in `from`'s return type.  There's simply no choice what you return.  It would be warning us to make sure our code builds, or warning us to not write bugs into our code.

If you assume they added (and emphasized) that warning for a reason, there must be more to it than that.  It must be more general than "don't do what you can't do".  The reason they warn us not to let it fail is because From is integral to Rust's failure handling mechanisms, and it's not at all clear what should happen when failure handling fails.

dunmatt: ... so... I think I just accidentally convinced myself that this PR should be rejected.  *sigh*

Aehmlo: Not to beat a dead horse, but I still believe that the meaning for that documentation for `From` is simply "the return value is not an error type." This is especially supported by the fact that [you can `impl From<A> for Option<B>` to mimic `TryFrom`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c1b0510ed7f7d091e19ab12ff7d8564e) (and so there is indeed some flexibility in the return type). Regardless, I do think that `TryFrom` is the better fit here. May I ask why `From` is necessary for your use case/why `TryFrom` is insufficient?

dunmatt: It's not that the trait is insufficient on its own (I mean, it's a bit cumbersome to have to do yet more error handling, but at the end of the day that's not that bid a deal), it's more that the difficulty of making the change is greater than the amount I care about it being made.  I've already sunk 100x more time into this PR than its actually worth to me, and if I were to double that investment I still don't think that's enough to adequately implement TryFrom.

Aehmlo: That makes sense, thanks. I think I might try to make up a `TryFrom` PR, fwiw. Thanks for your efforts on this.

PR #142, opened 2019-05-14
Add non-autoconvert impl of si::marker::impl_from!

### LINKED ISSUE #141: Build fails when autoconvert feature is disabled
The new `impl_from!` macro unconditionally uses `si::change_base`, which is only available when the `autoconvert` feature is enabled. Disabling the feature causes the following build error:
```
error[E0432]: unresolved import `si::change_base`
  --> src\si\mod.rs:93:14
   |
93 |     use si::{change_base, Dimension, Quantity, Units};
   |              ^^^^^^^^^^^ no `change_base` in `si`
```

This is a regression in uom 0.23.0, introduced in #132.

dmit:
Fixes #141

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ matrix:
     script: |
       set -e
       cargo test --all --verbose --features "use_serde"
+      cargo test --verbose --no-default-features --features "f32 si"
       cargo test --verbose --no-default-features --features "autoconvert f32 si use_serde"
       cargo test --verbose --no-run --no-default-features --features "autoconvert usize isize bigint bigrational si std use_serde"
       cargo test --verbose --no-run --no-default-features --features "autoconvert usize u8 u16 u32 u64 isize i8 i16 i32 i64 bigint biguint rational rational32 rational64 bigrational f32 f64 std use_serde"
--- a/src/si/mod.rs
+++ b/src/si/mod.rs
@@ -90,7 +90,7 @@ storage_types! {
 
 /// Primitive traits and types representing basic properties of types specific to the SI.
 pub mod marker {
-    use si::{change_base, Dimension, Quantity, Units};
+    use si::{Dimension, Quantity, Units};
     use Kind;
 
     /// AngleKind is a `Kind` for separating angular quantities from their indentically dimensioned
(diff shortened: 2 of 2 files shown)


iliekturtles:
Thanks for the PR! Forgot about auto convert. Would be good to also update the build script to catch this in the future.

I have limited internet access right now so I won't be able to finish reviewing/merge until this weekend at the earliest.

dmit:
Added a test run with `--no-default-features --features "f32 si"` to the Travis config. Is this what you had in mind?

Technically, `"f32"` is the minimum feature set needed to compile and run tests, but excluding `si` ignores a big chunk of them and doesn't catch the bug being fixed in this PR because the relevant code is skipped by the compiler.

iliekturtles:
Sorry for the long delay. Limited internet access turned out to be pretty much none!

Can you move the `#[feature(...)]` blocks on top of the `impl` instead of the `fn`. The auto convert impl should have two generic parameters `Ul` and `Ur` while the not(autoconvert) feature should only have a single parameter `U`. See an example at https://github.com/iliekturtles/uom/blob/457713283369d6768f662a51252090ba6ff40f86/src/si/thermodynamic_temperature.rs#L102-L124

dmit:
Done!

Side note: I've never seen libraries change their API signatures depending on enabled crate features before. Not sure how I feel about that. Are there any official guidelines? For some reason my expectation was that turning on features should be a strictly additive change - enabling new functionality without breaking existing code.

iliekturtles:
I think the signature change is still considered additive. When `autoconvert` is not set only quantities with the same base units can be converted with `From` (or other operations gated on `autoconvert`). e.g. Length stored as meters can't interact with length stored as kilometers. Enabling `autoconvert` adds conversions where the base units are not the same. e.g. length stored as meters and length stored as kilometers can now interact and no existing code breaks.

If you don't have any other changes and if you can squash the third commit into the first I'll merge and release `v0.23.1`!

dmit:
I'm not knowledgeable enough at type theory to tell whether adding a type parameter to an impl is additive in the mathematical sense, but your explanation sounds good to me. Don't think it matters in practice either - the code will still compile in 99% of cases.

Squashed the commits. :shipit:

iliekturtles:
Thanks so much for the PR and your patience while I was away. Working on getting v0.23.1 released right now.

iliekturtles:
v0.23.1 was just published on crates.io!

PR #144, opened 2019-06-02
Correct `Quantity::new` to be zero-cost for float storage types.

### LINKED ISSUE #143: Creation of basic quantities is not zero-cost
Compiling the following program produces an extra `0.0f + value` operation in every `Angle::new()` call:
```
use uom::si::{SI, angle::{Angle, radian}};

#[inline(never)]
fn half_way(new: Angle::<SI<f64>, f64>) -> f64 {
    new.get::<radian>().sin().sin()
}
    
#[inline(never)]
fn value(new: f64) -> f64 {
    new.sin()
}

fn main() {
    let b1 = value(11111.0);
    let b2 = value(22222.0);
    let c1 = half_way(Angle::<SI<f64>, f64>::new::<radian>(b1));
    let c2 = half_way(Angle::<SI<f64>, f64>::new::<radian>(b2));
    println!("{}", b1 + b2 + c1 + c2);
```
The pertaining LLVM-IR output is:
```
; call uom_test::value
  %0 = tail call fastcc double @_ZN8uom_test5value17hc4ec51985f4bef61E(double 1.111100e+04)
; call uom_test::value
  %1 = tail call fastcc double @_ZN8uom_test5value17hc4ec51985f4bef61E(double 2.222200e+04)
  %2 = fadd double %0, 0.000000e+00
; call uom_test::half_way
  %3 = tail call fastcc double @_ZN8uom_test8half_way17ha0a6a5c0bfbcb7a8E(double %2)
  %4 = fadd double %1, 0.000000e+00
; call uom_test::half_way
  %5 = tail call fastcc double @_ZN8uom_test8half_way17ha0a6a5c0bfbcb7a8E(double %4)
```
I even tried compiling with `-ffast-math` but had no success.

iliekturtles:
Use a default constant of `-0.0` to allow for floating point
optimizations. For a value, `v: Float`, adding `-0.0` is a no-op while
adding `0.0` will change the sign if `v` is `-0.0`. Resolves #143.

```
   v
 0.0 + -0.0 =  0.0
-0.0 +  0.0 =  0.0 # v + 0.0 != v
-0.0 + -0.0 = -0.0
```

@raimundomartins with the changes in this PR your example program generates the expected zero-cost LLVM-IR! If you could review / test with your full code that would be greatly appreciated.

```llvm
; call uom_opt_test::value
%0 = tail call fastcc double @_ZN12uom_opt_test5value17h5bf2d6ab643b092cE(double 1.111100e+04)
; call uom_opt_test::value
%1 = tail call fastcc double @_ZN12uom_opt_test5value17h5bf2d6ab643b092cE(double 2.222200e+04)
; call uom_opt_test::half_way
%2 = tail call fastcc double @_ZN12uom_opt_test8half_way17h905d0903778e0b79E(double %0)
; call uom_opt_test::half_way
%3 = tail call fastcc double @_ZN12uom_opt_test8half_way17h905d0903778e0b79E(double %1)
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -346,6 +346,32 @@ pub mod si;
 #[cfg(test)]
 mod tests;
 
+/// Operations performed on the constant portion of the [conversion factor][factor]. Used to help
+/// guide optimizations when floating point underlying storage types are used.
+///
+/// For a value, `v: Float`, adding `-0.0` is a no-op while adding `0.0` will change the sign if
+/// `v` is `-0.0`. The opposite is true for subtraction.
+///
+/// ```ignore
+///    v
+///  0.0 + -0.0 =  0.0
+/// -0.0 +  0.0 =  0.0 // v +  0.0 != v
+/// -0.0 + -0.0 = -0.0
+///  0.0 - -0.0 =  0.0
+/// -0.0 -  0.0 =  0.0
+/// -0.0 - -0.0 =  0.0 // v - -0.0 != v
+/// ```
+///
+/// [factor]: https://jcgm.bipm.org/vim/en/1.24.html
+#[derive(Clone, Copy, Debug)]
+pub enum ConstantOp {
+    /// Hint that the constant is being added to a value.
+    Add,
+
+    /// Hint that the constant is being subtracted from a value.
+    Sub,
+}
+
 /// Trait to identify [units][units] which have a [conversion factor][factor].
 ///
 /// [units]: http://jcgm.bipm.org/vim/en/1.13.html
@@ -355,9 +381,8 @@ pub trait Conversion<V> {
     type T: ConversionFactor<V>;
 
     /// Coefficient portion of [conversion factor][factor] for converting the given unit to the
-    /// base unit for the quantity: `(value * coefficient()) + constant()`.
-    ///
-    /// Default implementation returns `Self::T::one()`.
+    /// base unit for the quantity: `(value * coefficient()) + constant()`. Implementation should
+    /// return the multiplicative identity (`Self::T::one()`) if no coefficient exists.
     ///
     /// [factor]: https://jcgm.bipm.org/vim/en/1.24.html
     #[inline(always)]
@@ -366,13 +391,15 @@ pub trait Conversion<V> {
     }
 
     /// Constant portion of [conversion factor][factor] for converting the given unit to the base
-    /// unit for the quantity: `(value * coefficient()) + constant()`.
-    ///
-    /// Default implementation returns `Self::T::zero()`.
+    /// unit for the quantity: `(value * coefficient()) + constant()`. Implementation should return
+    /// the additive identity (`Self::T::zero()`) if no constant exists. See
+    /// [ConstantOp](enum.ConstantOp.html) documentation for details about parameter use to use to
+    /// ensure method optimizes correctly.
     ///
     /// [factor]: https://jcgm.bipm.org/vim/en/1.24.html
     #[inline(always)]
-    fn constant() -> Self::T {
+    #[allow(unused_variables)]
+    fn constant(op: ConstantOp) -> Self::T {
         <Self::T as num::Zero>::zero()
     }
 
@@ -433,6 +460,14 @@ storage_types! {
 
     impl ::Conversion<V> for V {
         type T = V;
+
+        #[inline(always)]
+        fn constant(op: ::ConstantOp) -> Self::T {
+            match op {
+                ::ConstantOp::Add => -<Self::T as ::num::Zero>::zero(),
+                ::ConstantOp::Sub => <Self::T as ::num::Zero>::zero(),
+            }
+        }
 
         #[inline(always)]
         fn into_conversion(&self) -> Self::T {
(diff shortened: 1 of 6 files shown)


raimundomartins:
My "full code" does not omit the add 0.0, but I'm unable to pinpoint why, since it's just the following benchmark tests:    
```
    #[bench]
    fn compare_uom_trigonometric_with_direct(b: &mut Bencher) {
        b.iter(||
            (0..1000).map(|i| i as f64).
            fold(0.0, |old, new| {
                old + new.sin().sin()
            })
        )
    }
    #[bench]
    fn uom_trigonometric(b: &mut Bencher) {
        use uom_ops::Trigonometric;
        b.iter(||
            (0..1000).map(|i| i as f64).
            fold(0.0, |old, new| {
                old + Angle::new::<radian>(new.sin()).sin().get::<ratio>()
            })
        )
    }
```
with uom_ops::Trigonometric being:
```
pub trait Trigonometric<Result> {
    fn sin(&self) -> Result;
    fn cos(&self) -> Result;
    fn tan(&self) -> Result;
}
macro_rules! impl_si_trait {
    {$trait:ident, ($input:ident, $input_unit:ty), ($output:ident, $output_unit:ty): [$($func:ident),+]$(, {$($extra:tt)*})?} => {
        impl<T> $trait<$output<SI<T>, T>> for $input<SI<T>, T>
        where
            T: uom::num::Float + uom::Conversion<T>,
            radian: uom::Conversion<T, T = <T as uom::Conversion<T>>::T>,
            ratio: uom::Conversion<T, T = <T as uom::Conversion<T>>::T>,
        {
            $(fn $func(&self) -> $output<SI<T>, T> {
                $output::<SI<T>, T>::new::<$output_unit>(self.get::<$input_unit>().$func())
            })+
            $($($extra)*)?
        }
    };
}

impl_si_trait!{Trigonometric, (Angle, radian), (Ratio, ratio): [sin, cos, tan]} }
```
I fail to see why it doesn't vanish here but vanishes in the previous example program. I did make sure I was `Compiling uom v0.23.1 (https://github.com/iliekturtles/uom/?rev=ea778679fa7e38156a97538dca48ab2a999f7a37#ea778679)`

Now a bit off-topic, would you be interested to collaborate on a crate which impl's functions similar to the above? Surely there's a better implementation for it than this, and I find these functions to be of such basic use to require users to extract a value, use the op, and convert to another value. (I actually believe it removes all value of the uom crate itself). If so, how do you wish to proceed, by opening an issue?

iliekturtles:
I'll checkout those benchmark examples this weekend and see if I can reproduce / solve the issue.

For `sin`, `cos`, `tan`, and other trigonometric functions I think it would be appropriate to put the functions directly into `uom` without the need to rely on a separate crate. I haven't had a chance to add a new issue for this to list out all the functions to implement but will do so this weekend if you don't get to it first. See `length` where `hypot()` is explicitly implemented. https://github.com/iliekturtles/uom/blob/c02d3065977270d8ceeb00be6dc3017b272d4827/src/si/length.rs#L67-L82

iliekturtles:
I ended up being busy all weekend and will try to review this week.

iliekturtles:
I found the issue. `to_base` does `+ N::constant()` so adding `-0.0` becomes a no-op. `from_base` does `- N::constant()` and all of of a sudden that `-0.0` becomes `0.0` and is no longer a no-op! All the original PR did was move the issue from `to_base` to `from_base`!

https://github.com/iliekturtles/uom/blob/c02d3065977270d8ceeb00be6dc3017b272d4827/src/system.rs#L284-L286
https://github.com/iliekturtles/uom/blob/c02d3065977270d8ceeb00be6dc3017b272d4827/src/system.rs#L302-L304

So far I have two thoughts about how to fix the issue. Neither are great so I welcome suggestions.
 1. Change `constant()` into two functions: `additive_constant` and `subtractive_constant`. Each version can return an appropriately signed `0.0`.
 1. Change `constant()` to take a parameter suggesting how the constant will be used. The function can now return the appropriate constant.

@raimundomartins I've currently gone with the second option. The updated code causes llvm to optimize the benchmarks down to a single llvm-ir function definition. If you could review the code and double check that the optimization is still happening that would be greatly appreciated.

raimundomartins:
I had thought of that, but some quick tests didn't point to the opposite operation to not being optimized, so I thought that `- -0.0` was being used. Anyway, I'm glad to report that in my case these operations are zero-cost!
I just thought I should mention that the default impl of `Conversion<T>` should probably get the match arms as well.
https://github.com/iliekturtles/uom/blob/32d36e24d81695c442d1ebd3db15912e3a39f72f/src/lib.rs#L404-L406

iliekturtles:
Glad to hear it's fixed this time!

For the default implementation of `fn constant` I had started doing that last night but was a bit rushed and didn't pursue it. I was concerned how it would interact when T was an unsigned type. I'm not sure if the default implementation is actually used anywhere so I may just remove it. I'll merge the PR as soon as I have that resolved.

iliekturtles:
Merged! I'll try to get a new version published in the next couple days!

iliekturtles:
[v0.24.0](https://github.com/iliekturtles/uom/releases/tag/v0.24.0) published! Thanks again for submitting the issue.

PR #146, opened 2019-06-13
Make Quantity repr(transparent)

### LINKED ISSUE #145: Quantity should be `repr(transparent)`
Is there a reason why this is not the case ? Without `repr(transparent)` `Quantity` has a struct ABI, instead of the ABI of `T` (scalar float or scalar integer). This means that one cannot use `Quantity` in C FFI, and it also might mean that the compiler might generate sub-optimal code for those situations in which the struct ABI is worse than the float or integer one.

gnzlbg:
Closes #145 .

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -233,6 +233,7 @@ macro_rules! system {
         /// ```
         ///
         /// * <http://jcgm.bipm.org/vim/en/1.1.html>
+        #[repr(transparent)]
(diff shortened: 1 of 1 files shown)


gnzlbg:
This would require bumping the minimum supported Rust version to something more recent.

iliekturtles:
Merged. Thanks!

PR #148, opened 2019-06-13
Increase minimum supported `rustc` version to 1.28.0.

iliekturtles:
Changed in anticipation of adding `#[repr(transparent)]` in #145.
Zero-cost codegen note regarding `rustc` 1.25.0 is now no longer
relevant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ cache:
 sudo: false
 
 rust:
-- 1.24.1    # Oldest supported
+- 1.28.0    # Oldest supported
 - stable
 - beta
 - nightly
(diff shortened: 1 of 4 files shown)


iliekturtles:
Build failing for 1.28.0 with the following error:

```
error: failed to parse manifest at `D:\Code\Test\rust\uom\uom-macros\Cargo.toml`

Caused by:
  editions are unstable

Caused by:
  feature `edition` is required

this Cargo does not support nightly features, but if you
switch to nightly channel you can add
`cargo-features = ["edition"]` to enable this feature
```

`edition` keyword is now recognized by `cargo` but not yet stabilized. Previously builds for 1.24.1 would still pass because the `edition` keyword was not recognized and hence ignored by `cargo`. May need to do a 1.28.0 / 1.31.0 split.

PR #152, opened 2019-06-20
Fix macros in features.rs

Aehmlo:
The macros as documented should use `all`, not `any`, unless I'm missing something.


iliekturtles:
The code is correct and the documentation is unclear. Should be "autoconvert or test."

Aehmlo:
That would explain why all the tests broke!

Issue #154, opened 2019-06-24
Support references to other units in a unit's conversion factor


iliekturtles:
Support references to other units in a unit's conversion factor.

via #31 

For unit aliases it would be cool if the following were possible:

    @bit: prefix!(none) / 8.0; "b", "bit", "bits";
    @shannon: @bit; "Sh", "shannon", "shannons";

And for conversions:

    @deciban: prefix!(deci) * @hartley; "dHart", "deciban", "decibans";

Lukazoid:
I have found I am able to do this:

    @shannon: <super::bit as crate::Conversion<Self::T>>::coefficient(); "Sh", "shannon", "shannons";

And

    @deciban: prefix!(deci) * <super::hartley as crate::Conversion<Self::T>>::coefficient(); "dHart", "deciban", "decibans";

However it's not the most concise, perhaps a macro to generate the equivalent?

iliekturtles:
Updating the `@coefficient` and `@constant` arms of the `quantity!` macro was what I was first thinking. I've had troubles replacing a single token (e.g. `@bit`) in an arbitrary expression in the past but that is where to start looking.

https://github.com/iliekturtles/uom/blob/0dd06308484ea415c87b0b762004286966e3fc3e/src/quantity.rs#L497-L505

Alternatively, and I just thought of this while writing my response, would be to do the expansion with a totally separate macro (or arm of an existing macro):

```rust
@deciban: prefix!(deci) * unit_macro_or_arm_in_existing_macro!(hartley); "dHart", "deciban", "decibans";
```

iliekturtles:
For reference this didn't end up getting completed as part of #158. There were issues with some underlying storage types such as `BigInt`. See the error message below along with the code for the two macros.

```sh
error[E0277]: the trait bound `si::information_rate::bit_per_second: Conversion<num_rational::Ratio<num_
bigint::BigInt>>` is not satisfied
  --> src\unit_coef.rs:6:9
   |
6  |         <super::$other_unit as $crate::Conversion<Self::T>>::coefficient()
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Conversion<num_
rational::Ratio<num_bigint::BigInt>>` is not implemented for `si::information_rate::bit_per_second`
   |
  ::: src\si\information_rate.rs:53:46
   |
[...]

Issue #157, opened 2019-07-15
Build failure after patch release of atty v0.2.12.


iliekturtles:
https://travis-ci.org/iliekturtles/uom/jobs/558710453#L465

A recent patch release of `atty` from v0.2.11 to v0.2.12 upgraded the crate to Rust 2018. This causes build failures for `rustc` 1.28.0. `atty` isn't a direct dependency of `uom`. What can be done in `uom` to resolve this issue? Can the direct `uom` dependency be specified with enough detail in order to get the older version of `atty`?

iliekturtles:
https://github.com/softprops/atty/issues/36

softprops was kind enough to release v0.2.13 which no longer requires the 2018 edition.

PR #158, opened 2019-07-15
Information quantities

### LINKED ISSUE #31: Add support for bytes?
Would support for 'bytes' be within scope for this library? Would that work out well with using f32 / f64 storage? It seems like perhaps using `u64` for bytes, but floating point for all larger units might be useful?

Lukazoid:
Closes #31, #154

### DIFF
--- a/src/si/information.rs
+++ b/src/si/information.rs
@@ -1,0 +1,80 @@
+//! Information (dimensionless quantity).
+
+quantity! {
+    /// Information (dimensionless quantity).
+    quantity: Information; "information";
+    /// Dimension of information, 1 (dimensionless).
+    dimension: ISQ<
+        Z0,     // length
+        Z0,     // mass
+        Z0,     // time
+        Z0,     // electric current
+        Z0,     // thermodynamic temperature
+        Z0,     // amount of substance
+        Z0>;    // luminous intensity
+    kind: dyn(::si::marker::InformationKind);
+    units {
+        // Base-2.
+        @yobibit: prefix!(yobi) / 8.0; "Yib", "yobibit", "yobibits";
+        @yottabit: prefix!(yotta) / 8.0; "Yb", "yottabit", "yottabits";
+        @zebibit: prefix!(zebi) / 8.0; "Zib", "zebibit", "zebibits";
+        @zettabit: prefix!(zetta) / 8.0; "Zb", "zettabit", "zettabits";
+        @exbibit: prefix!(exbi) / 8.0; "Eib", "exbibit", "exbibits";
+        @exabit: prefix!(exa) / 8.0; "Eb", "exabit", "exabits";
+        @pebibit: prefix!(pebi) / 8.0; "Pib", "pebibit", "pebibits";
+        @petabit: prefix!(peta) / 8.0; "Pb", "petabit", "petabits";
+        @tebibit: prefix!(tebi) / 8.0; "Tib", "tibibit", "tibibits";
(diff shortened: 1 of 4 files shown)


iliekturtles:
Awesome! Thanks for the PR. I'll start reviewing this week. You can ignore the 1.28.0 build failures for the moment. See #157.

Lukazoid:
Will fix the fmt and clippy errors tomorrow.

Lukazoid:
Any idea why clippy fails with warnings around my use of `InformationKind` but the use of `AngleKind` doesn't appear to? I can't figure out what's different.

iliekturtles:
You just need the `dyn` keyword and some parenthesis. I recently enabled the lint that requires this.

https://github.com/iliekturtles/uom/blob/c5dc8b43b17051e3ecc3c05140451d8d935b6475/src/si/angle.rs#L15

iliekturtles:
First pass at a review is complete. I still need to go through and review each of the details to validate conversion factors and descriptions.

 * Rebase your changes on top of the latest `master` to get a number of fixes. The issue with the 1.28.0 build failures have been resolved (you may need to `cargo update`) locally. There are a couple uses of `crate::` in these changes that need to be resolved.
 * `unit_coef` and `unit_const` should be able to be moved out of the `quantity!` macro. I don't think there is a need to generate these for each quantity.
 * information.rs
   * Remove the blank line 17 and 55. Remove trailing whitespace from 53.
   * Should the abbreviation for `deciban` be `dHart`?
 * information_rate.rs
    * Change line 6 to reference "information rate" instead of "baud rate." Perhaps note in one of the doc comments the equivalence?
    * Wrap 45 and 46 at 100 characters.
    * Add test cases for all the information rate units.
 * mod.rs
    * Remove the blank line 284.
 * prefix.rs
    * Note that SI and binary prefixes are implemented in the macro's doc comment.
    * Change line 36 to `// Binary prefixes.`

Lukazoid:
@iliekturtles Thanks, I will get on to those changes this evening after work 👍

Lukazoid:
Seems my macros don't work for the bigint storage types, will try and figure out a solution.

iliekturtles:
I'll take a look tomorrow as well. It may be a limitation of how the `quantity!` macro is setup so don't put too much time into it.

iliekturtles:
Conversion factors for some underlying storage types are currently (I want to change this in v2.0.0) first converted into `f64` and then into the resulting type such as `Ratio<BigInt>`. Using the conversion factor of another unit returns the resulting the resulting type while the `quantity!` internals are still working as `f64`.

Using the conversion factor for the unit as `f64` may fix the error but may also introduce some precision issues. Give it a try and we can't get that working I think it's OK to table this for the time being and revisit after v1.0.0 is released.

iliekturtles:
I did another round of review and pushed changes to a branch in my repo: https://github.com/Lukazoid/uom/compare/information_quantities...iliekturtles:Lukazoid-information_quantities. If you want to review/apply to your branch/squash I think we can merge!

 * Fixed `mebi` prefix (was `mibi`).
 * Re-ordered units from large to small to match conventions of other quantities.
 * Ended up removing the `unit_coef!` and `unit_const!` macros. Issues with other underlying storage types like `BigInt` aren't worth holding up the rest of this PR.
 * Cleaned up some comments.

Lukazoid:
Thanks @iliekturtles I have done this.

iliekturtles:
Awesome! I'm planning do a final review tonight and then merge.

PR #160, opened 2019-08-12
Added AngularJerk.

dunmatt:
Hello again!  I found myself needing even higher order angular units so I figured I'd share with the class.  Should be a pretty easy review, it's mostly a copy pasta of AngularAcceleration.


iliekturtles:
Thanks so much for the PR! It was an easy one.

PR #169, opened 2019-12-07
chore: bump deps

tim77:


iliekturtles:
The `static_assert` bump requires fixing the macro calls: https://travis-ci.org/iliekturtles/uom/jobs/622083216#L246

tim77:
Oh right. Before i can't understand why tests not passed. Should i open a new issue about this? I would really like to send PR with fix but this is above my skill for now. Thank you.

iliekturtles:
Check out the `static_assertions` documentation: https://docs.rs/static_assertions/1.1.0/static_assertions/macro.assert_impl_all.html. The label (`arguments_impl` in the example below) is removed and the type is separated from the traits with a colon.

https://github.com/iliekturtles/uom/blob/3a236e542b0abe5bd19f0dc46465490a64fb50a2/src/tests/asserts.rs#L6-L7

changes to

```rust
 assert_impl_all!(Arguments<Q<Z0, Z0, Z0>, meter>: Clone, Copy); 
```

iliekturtles:
I just did some more digging and unfortunately neither `quickcheck` nor `static_assertions` can be upgraded without an accompanying MSRV (minimum supported Rust version) update for `uom` at the same time. I'm closing the PR since I don't want to bump the MSRV yet. Thank you for the PR!

PR #170, opened 2019-12-11
chore: Update to static_assertions to 1

eclipseo:
Signed-off-by: Robert-André Mauchin <zebob.m@gmail.com>

PR to replace #169

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ typenum = "1.9"
 approx = "0.3"
 quickcheck = "0.8.5"
 serde_json = "1.0"
-static_assertions = "^0.3.4"
+static_assertions = "1"
 
 [features]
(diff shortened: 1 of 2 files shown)


eclipseo:
This bump MSRV to 1.37 at least so you may want to delay that.

iliekturtles:
Yeah, I'd like to delay bumping the MSRV. Is it possible to use more recent versions of either dependencies on 1.31?

eclipseo:
Quickcheck is good, it's static_assertions > 1 that needs 1.37.

iliekturtles:
I just did some more digging and unfortunately neither `quickcheck` nor `static_assertions` can be upgraded without an accompanying MSRV (minimum supported Rust version) update for `uom` at the same time. I'm closing the PR since I don't want to bump the MSRV yet. Thank you for the PR!

PR #171, opened 2019-12-11
chore: Bump quickcheck to 0.9

eclipseo:
Signed-off-by: Robert-André Mauchin <zebob.m@gmail.com>

Split from #170

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ typenum = "1.9"
 
 [dev-dependencies]
 approx = "0.3"
-quickcheck = "0.8.5"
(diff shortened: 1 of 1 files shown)


iliekturtles:
I just did some more digging and unfortunately neither `quickcheck` nor `static_assertions` can be upgraded without an accompanying MSRV (minimum supported Rust version) update for `uom` at the same time. I'm closing the PR since I don't want to bump the MSRV yet.

iliekturtles:
EDIT: Thanks so much for the PR. As soon as the MSRV is bumped I'll make sure these dependencies also get updated.

PR #175, opened 2020-02-11
Add Curvature.

jhelovuo:


src/si/curvature.rs
```diff
@@ -0,0 +1,49 @@
...
+        N1,     // length
+        Z0,     // mass
+        Z0,     // time
+        Z0,     // electric current
+        Z0,     // thermodynamic temperature
+        Z0,     // amount of substance
+        Z0>;    // luminous intensity
+    kind: dyn (::si::marker::AngleKind);
+    units {
+        @radian_per_meter: prefix!(none); "rad/s", "radian per second", "radians per second";
```
iliekturtles: Abbreviation and descriptions should be per meter.

jhelovuo:
Hi!
Thanks for the feedback. It seems I forgot to commit the correction for unit names. Now they should be in.

I also added some alternative units as you suggested. The selection of units is based on some googling. The only units I could find used anywhere were rad/m and deg/mm (for characterising wool fibres!). Curvature is important in larger scale engineering also, such as road and railway construction, but they seem not to use curvature quantity directly, but specify radius of curvature (Length) or degree of curvature (Angle) instead.

Tests of millimeter units are commented out, because they fail due to a numerical precision issue:

> assertion failed: `(left == right)`
>  left: `1000.0`,
> right: `999.99994`

I do not know how this should be handled in uom library.

iliekturtles:
Everything looks really good! I found a few minor comment issues and got the tests running by using `assert_approx_eq`. If you could apply the diff below and then squash your changes I will merge!

```diff
diff --git a/src/si/curvature.rs b/src/si/curvature.rs
index d9f923c..689fc7a 100644
--- a/src/si/curvature.rs
+++ b/src/si/curvature.rs
@@ -1,9 +1,9 @@
-//! [Curvature](https://en.wikipedia.org/wiki/Curvature) (base unit radians per meter, rad · m⁻¹).
+//! [Curvature](https://en.wikipedia.org/wiki/Curvature) (base unit radian per meter, m⁻¹).
 
 quantity! {
-    /// Curvature (base unit radians per meter, rad · m⁻¹).
+    /// Curvature (base unit radian per meter, m⁻¹).
     quantity: Curvature; "curvature";
-    /// Dimension of Curvature, L⁻¹ (base unit radians per meter, rad · m⁻¹).
+    /// Dimension of curvature, L⁻¹ (base unit radian per meter, m⁻¹).
     dimension: ISQ<
         N1,     // length
         Z0,     // mass
@@ -45,14 +45,13 @@ mod tests {
         fn check_units() {
             test::<a::radian, l::meter, c::radian_per_meter>();
             test::<a::degree, l::meter, c::degree_per_meter>();
-            //test::<a::radian, l::millimeter, c::radian_per_millimeter>();
-            //test::<a::degree, l::millimeter, c::degree_per_millimeter>();
+            test::<a::radian, l::millimeter, c::radian_per_millimeter>();
+            test::<a::degree, l::millimeter, c::degree_per_millimeter>();
 
             fn test<A: a::Conversion<V>, L: l::Conversion<V>, C: c::Conversion<V>>() {
-                Test::assert_eq(&Curvature::new::<C>(V::one()),
-                    &( Angle::new::<A>(V::one()) / Length::new::<L>(V::one()) ).into() );
+                Test::assert_approx_eq(&Curvature::new::<C>(V::one()),
+                    &(Angle::new::<A>(V::one()) / Length::new::<L>(V::one())).into());
             }
-
         }
     }
 }
```

jhelovuo:
Ok. Patch apply and squash are now done and pushed.

iliekturtles:
Merged! Thanks so much for the PR.

PR #182, opened 2020-04-23
Add tropical and sidereal times

Aehmlo:


Aehmlo:
I haven't found much consistency in the abbreviations used, so I've gone ahead and changed it to the parenthetical style, since that seems like a good compromise.

iliekturtles:
Thanks again! Merged.

PR #183, opened 2020-05-05
Fix a misplaced doc-comment

adamreichold:


iliekturtles:
Merged! Thanks so much for the PR.

adamreichold:
> Merged! Thanks so much for the PR.

I understand that the change is rather inconsequential, but rustc was emitting a warning for each line of that doc-comment which makes even a check build quite noisy.

PR #184, opened 2020-05-05
Implement inverse trigonometric functions from ratio to angle.

### LINKED ISSUE #168: Feature requrest: Need asin, acos, atan{,2} support on length
I had some work on Cartesian to spherical coordinate conversion with `uom`.
However, it lacks support on inverse trigonometric functions, and thus I have to use `get()` as workaround. Is it planned to add `asin`, `acos`, `atan` and `atan2` functions?

adamreichold:
Fixes #168

### DIFF
--- a/src/si/ratio.rs
+++ b/src/si/ratio.rs
@@ -1,5 +1,8 @@
 //! Ratio (dimensionless quantity).
 
+#[cfg(feature = "std")]
+use super::angle::{Angle, radian};
+
 quantity! {
     /// Ratio (dimensionless quantity).
     quantity: Ratio; "ratio";
@@ -26,6 +29,51 @@ quantity! {
         @part_per_billion: 1.0_E-9; "ppb", "part per billion", "parts per billion";
         @part_per_trillion: 1.0_E-12; "ppt", "part per trillion", "parts per trillion";
         @part_per_quadrillion: 1.0_E-15; "ppq", "part per quadrillion", "parts per quadrillion";
+    }
+}
+
+/// Implementation of various stdlib inverse trigonometric functions
+#[cfg(feature = "std")]
+impl<U, V> Ratio<U, V>
+where
+    U: ::si::Units<V> + ?Sized,
+    V: ::num::Float + ::Conversion<V>,
+    radian: ::Conversion<V, T = V::T>,
+{
+    /// Computes the value of the inverse cosine of the ratio.
+    #[inline(always)]
+    pub fn acos(self) -> Angle<U, V> {
+        Angle::new::<radian>(self.value.acos())
+    }
+
+    /// Computes the value of the inverse hyperbolic cosine of the ratio.
+    #[inline(always)]
+    pub fn acosh(self) -> Angle<U, V> {
+        Angle::new::<radian>(self.value.acosh())
+    }
+
+    /// Computes the value of the inverse sine of the ratio.
+    #[inline(always)]
+    pub fn asin(self) -> Angle<U, V> {
+        Angle::new::<radian>(self.value.asin())
+    }
+
+    /// Computes the value of the inverse hyperbolic sine of the ratio.
+    #[inline(always)]
+    pub fn asinh(self) -> Angle<U, V> {
+        Angle::new::<radian>(self.value.asinh())
+    }
+
+    /// Computes the value of the inverse tangent of the ratio.
+    #[inline(always)]
+    pub fn atan(self) -> Angle<U, V> {
+        Angle::new::<radian>(self.value.atan())
+    }
+
+    /// Computes the value of the inverse hyperbolic tangent of the ratio.
+    #[inline(always)]
+    pub fn atanh(self) -> Angle<U, V> {
+        Angle::new::<radian>(self.value.atanh())
     }
 }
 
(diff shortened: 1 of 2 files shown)


adamreichold:
@iliekturtles I am somewhat at a loss as to the correct usage of the `Convert` trait, i.e. I get
```
error[E0271]: type mismatch resolving `<V as Conversion<V>>::T == <si::angle::radian as Conversion<V>>::T`
  --> src/si/ratio.rs:76:9
   |
37 | impl<U, V> Ratio<U, V>
   |         - this type parameter
...
76 |         Angle::new::<radian>(self.value.atanh())
   |         ^^^^^^^^^^^^^^^^^^^^ expected type parameter `V`, found struct `si::angle::radian`
   |
   = note: expected associated type `<V as Conversion<V>>::T`
              found associated type `<si::angle::radian as Conversion<V>>::T`
```

Could you give me a hint?

src/si/ratio.rs
```diff
@@ -29,6 +32,51 @@ quantity! {
     }
 }
 
+/// Implementation of various stdlib inverse trigonometric functions
+#[cfg(feature = "std")]
+impl<U, V> Ratio<U, V>
+where
+    U: ::si::Units<V> + ?Sized,
+    V: ::num::Float + ::Conversion<V>,
+    radian: ::Conversion<V>,
```
iliekturtles, line 41: Adding `T = V::T` to the constraint should do the trick. See time.rs for an example.

https://github.com/iliekturtles/uom/blob/4fceafc98722464094aa7b658a8ff52d43944d00/src/si/time.rs#L89

adamreichold: Thank you! Works like a charm.

adamreichold:
@iliekturtles I wonder how to handle [`atan2`](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.atan2). Personally, I only it for `Length` values (to switch between coordinate systems), but strictly speaking it should apply to all pairs of quantities. Should I propose an implementation on `si::Length` or on the general `Quantity`?

iliekturtles:
I think I'm inclined towards implementing for all quantities but I'm second guessing myself as I write this sentence. Do you have any examples where the function would be used on non-length quantities?

adamreichold:
> Do you have any examples where the function would be used on non-length quantities?

After some digging, I found [a paper](https://www.researchgate.net/publication/303554491_DETERMINATION_OF_ELECTROMAGNETIC_PARAMETERS_AND_PHASE_RELATIONS_IN_TURBO-GENERATORS_BY_THE_AUTOMATED_CALCULATION_OF_THE_MAGNETIC_FIELD_IN_THE_SOFTWARE_ENVIRONMENT_FEMM) mentioning `atan2` when computing magnetic field strengths (of rotationally symmetric sources) where the computation a field quantity expansion involves `atan2`.

I guess this the basic example, any field which has a spherical or cylindrical symmetric might have modes which imply usage of `atan2` but it is admittedly a bit fringe...

Aehmlo:
An example of using the arctangent non-length quantities which immediately comes to mind for me is [the phase delay in viscoelastic solids](https://en.wikipedia.org/wiki/Dynamic_modulus#Ratio_between_loss_and_storage_modulus), which is something I could see myself wanting to work with in `uom` in the future.

adamreichold:
> I think I'm inclined towards implementing for all quantities

> An example of using the arctangent non-length quantities which immediately comes to mind

Amended the generic implementation on all `si::Quantity` values returning an `si::Angle`.

@iliekturtles Should be ready for review now.

iliekturtles:
Changes look great! If you can amend the commit with tests similar to what is done for `hypot` in `length.rs` I'll merge!

https://github.com/iliekturtles/uom/blob/4fceafc98722464094aa7b658a8ff52d43944d00/src/si/length.rs#L85-L102

adamreichold:
>  If you can amend the commit with tests

Added property-based tests for all new functions.

iliekturtles:
Clippy doesn't like `test_nan_or_eq`'s parameters. Since the function is only going to be called for floats the references can be dropped.

```diff
diff --git a/src/si/ratio.rs b/src/si/ratio.rs
index 8ceee33..4a3f853 100644
--- a/src/si/ratio.rs
+++ b/src/si/ratio.rs
@@ -161,33 +161,33 @@ mod tests {
             use si::quantities::*;
             use tests::Test;

-            fn test_nan_or_eq(yl: &V, yr: &V) -> bool {
-                (yl.is_nan() && yr.is_nan()) || Test::eq(yl, yr)
+            fn test_nan_or_eq(yl: V, yr: V) -> bool {
+                (yl.is_nan() && yr.is_nan()) || Test::eq(&yl, &yr)
             }

             quickcheck! {
                 fn acos(x: V) -> bool {
-                    test_nan_or_eq(&x.acos(), &Ratio::from(x).acos().get::<a::radian>())
+                    test_nan_or_eq(x.acos(), Ratio::from(x).acos().get::<a::radian>())
                 }

                 fn acosh(x: V) -> bool {
-                    test_nan_or_eq(&x.acosh(), &Ratio::from(x).acosh().get::<a::radian>())
+                    test_nan_or_eq(x.acosh(), Ratio::from(x).acosh().get::<a::radian>())
                 }

                 fn asin(x: V) -> bool {
-                    test_nan_or_eq(&x.asin(), &Ratio::from(x).asin().get::<a::radian>())
+                    test_nan_or_eq(x.asin(), Ratio::from(x).asin().get::<a::radian>())
                 }

                 fn asinh(x: V) -> bool {
-                    test_nan_or_eq(&x.asinh(), &Ratio::from(x).asinh().get::<a::radian>())
+                    test_nan_or_eq(x.asinh(), Ratio::from(x).asinh().get::<a::radian>())
                 }

                 fn atan(x: V) -> bool {
-                    test_nan_or_eq(&x.atan(), &Ratio::from(x).atan().get::<a::radian>())
+                    test_nan_or_eq(x.atan(), Ratio::from(x).atan().get::<a::radian>())
                 }

                 fn atanh(x: V) -> bool {
-                    test_nan_or_eq(&x.atanh(), &Ratio::from(x).atanh().get::<a::radian>())
+                    test_nan_or_eq(x.atanh(), Ratio::from(x).atanh().get::<a::radian>())
                 }
             }
         }
```

PR #187, opened 2020-05-08
Change trigonometric functions to return ratio instead of bare values

### LINKED ISSUE #185: Change trigonometric functions to return Ratio?
See https://github.com/iliekturtles/uom/issues/168#issuecomment-552442990. Should the existing trigonometric functions on `Angle` be changed to return `Ratio` instead of the underlying storage type?

adamreichold:
Fixes #185

### DIFF
--- a/src/si/angle.rs
+++ b/src/si/angle.rs
@@ -1,5 +1,8 @@
 //! Angle (dimensionless quantity).
 
+#[cfg(feature = "std")]
+use super::ratio::Ratio;
+
 quantity! {
     /// Angle (dimensionless quantity).
     quantity: Angle; "angle";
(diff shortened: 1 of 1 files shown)


adamreichold:
@iliekturtles While doing this, I was a bit surprised by the test
```
#[test]
fn from() {
    let r1: Angle<V> = Angle::<V>::from(V::one());
    let r2: Angle<V> = V::one().into();
    let _: V = V::from(r1);
    let _: V = r2.into();
}
```
for `Angle`. Shouldn't these kind of transformations be limited to `Ratio` as the above basically assumes `radian` (resp. the base unit)?

iliekturtles:
Thanks for the PR! Initial review from my phone looks good. I'll plan on merging once I can do a quick review from a computer.

For the `From`/`Into` impls I can't remember the history off the top of my head. Once I'm back at a computer I'll review the history and consider removing.

iliekturtles:
Merged! Thanks for your patience. I'll look at your other PR next.

For the `From`/`Into` impls they were included with the original PR to add `Angle`. There was no discussion about them and perhaps they were copied from the `Ratio` implementation. I've add #188 to discuss removing them.

PR #191, opened 2020-06-15
Add constants to Angle representing half and full turns

adamreichold:
Various trigonometric computations require Angle constants of these values but as construction via Angle::new is not possible in constants, this uses fixed storage types and base units to directly construct Quantity literals to wrap the standard libraries constants for π.

### DIFF
--- a/src/si/angle.rs
+++ b/src/si/angle.rs
@@ -29,6 +29,40 @@ quantity! {
     }
 }
 
+#[cfg(all(feature = "std", feature = "f32"))]
+impl Angle<::si::SI<f32>, f32> {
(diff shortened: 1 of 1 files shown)


iliekturtles:
Thanks for the PR! I meant to review today but got tied up. Will get done tomorrow morning. There was, possibly spurious, build failure so I restarted Travis.

iliekturtles:
Looks like the build failure is something introduced in Beta 1.45/Nightly and is unrelated to this PR so I have merged! I'm still investigating the failure and will report it upstream. Thanks again for the PR!

PR #193, opened 2020-06-18
Use `f{32|64}::powi` instead of `Typenum::Pow::powi`.

### LINKED ISSUE #192: Regression in powi on Windows beta/nightly
https://travis-ci.org/github/iliekturtles/uom/builds/698980459?utm_medium=notification&utm_source=email

```rust
extern crate typenum; // 2.12.0

fn main() {
    let v = 12.046623229980468750000000000000_f32;
    let a = a(v);
    let b = b(v);

    println!("std powi {:.30?}", a);
    println!("typenum  {:.30?}", b);
}

#[inline(never)]
fn a(v: f32) -> f32 {
    v.powi(3)
}

#[inline(never)]
fn b(v: f32) -> f32 {
    typenum::Pow::powi(v, typenum::P3::new())
}
```

```sh
/d/Code/Test/rust/tst (master)$ cargo run; rustc --version
   Compiling tst v0.1.0 (D:\Code\Test\rust\tst)
    Finished dev [unoptimized + debuginfo] target(s) in 3.90s
     Running `target\debug\tst.exe`
std powi 1748.219482421875000000000000000000
typenum  1748.219482421875000000000000000000
rustc 1.44.0 (49cae5576 2020-06-01)
/d/Code/Test/rust/tst (master)$ cargo +beta run; rustc +beta --version
    Finished dev [unoptimized + debuginfo] target(s) in 2.79s
     Running `target\debug\tst.exe`
std powi 1748.219604492187500000000000000000
typenum  1748.219482421875000000000000000000
rustc 1.45.0-beta.3359 (b7dc83a3f 2020-06-03)
```

iliekturtles:
In Rust 1.45 the upgrade to LLVM 10.0 changes the behavior of
`f{32|64}::powi` on Windows. This commit changes the `Quantity::powi`
implementation to simplify down to `f{32|64}::powi` instead of using
`Typenum::Pow::powi` which uses a less precise algorithm. Resolves #192.

Reviews welcome. Will plan to merge in a couple days.

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -797,6 +797,7 @@ macro_rules! system {
                 }
             }
 
+            std! {
             /// Raises a quantity to an integer power.
             ///
             #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
@@ -810,22 +811,21 @@ macro_rules! system {
             /// * `E`: `typenum::Integer` power.
             #[inline(always)]
             pub fn powi<E>(
-                self, e: E
+                self, _e: E
             ) -> Quantity<$quantities<$($crate::typenum::Prod<D::$symbol, E>),+>, U, V>
             where
                 $(D::$symbol: $crate::lib::ops::Mul<E>,)+
                 D::Kind: $crate::marker::Mul,
                 E: $crate::typenum::Integer,
-                V: $crate::typenum::Pow<E, Output = V> + $crate::Conversion<V>,
+                V: $crate::num::Float,
             {
                 Quantity {
                     dimension: $crate::lib::marker::PhantomData,
                     units: $crate::lib::marker::PhantomData,
-                    value: $crate::typenum::Pow::powi(self.value, e),
+                    value: self.value.powi(E::to_i32()),
                 }
             }
 
-            std! {
             /// Takes the square root of a number. Returns `NAN` if `self` is a negative
             /// number.
             ///


adamreichold:
I understand this is the faster solution, but I do wonder whether comments like https://docs.rs/typenum/1.12.0/src/typenum/type_operators.rs.html#103 still apply. Meaning that we should probably fix this in the `typenum` crate by making them forward to `{f32,f64}::powi` instead of doing it here (and for everywhere else `typenum` is used)?

adamreichold:
> I understand this is the faster solution, but I do wonder whether comments like https://docs.rs/typenum/1.12.0/src/typenum/type_operators.rs.html#103 still apply.

To answer my own question, it does seem to apply indeed, i.e. `{f32,f64}::powi` are not available in `no_std` environments and `core::instrinsics::powi{f32,f64}` are unstable. So this does look the best approach. Sorry for the noise...

src/system.rs
```diff
@@ -810,18 +810,20 @@ macro_rules! system {
...
-                V: $crate::typenum::Pow<E, Output = V> + $crate::Conversion<V>,
+                V: $crate::Conversion<V>,
             {
+                use $crate::ConversionFactor;
+
                 Quantity {
                     dimension: $crate::lib::marker::PhantomData,
                     units: $crate::lib::marker::PhantomData,
-                    value: $crate::typenum::Pow::powi(self.value, e),
+                    value: self.value.into_conversion().powi(E::to_i32()).value(),
```
adamreichold, line 826: This is probably just a lack of understanding, but looking at e.g. the implementation of `recip`, why is the trait bound
```
V: $crate::num::Float
```
and the implementation
```
value: self.value.powi(E::to_i32()),
```
not applicable here?

iliekturtles: I thought I test this when I did the original fix, but I must have been pretty distracted because I checked it again and it works. I just pushed the fix and am waiting on the build to finish. Thanks for the review!

PR #194, opened 2020-06-22
Added foot-candle unit to luminance

AnickaBurova:
Hi,

foot-candle is kind of a standard in the USA. Would it be possible to add it please.

I hope my code is doing all what is required. In case of any problems let me know, happy to adjust.


iliekturtles:
Thanks so much for the PR. Everything looks good and I have merged!

PR #196, opened 2020-07-18
Add support for solid angles mostly based on what we do for angles.

### LINKED ISSUE #190: Add support for solid angles and steradian
While the SI system supports angles measured in e.g. radians, it currently does not support solid angles measured in e.g. steradians.

adamreichold:
Closes #190

### DIFF
--- a/src/si/solid_angle.rs
+++ b/src/si/solid_angle.rs
@@ -1,0 +1,114 @@
+//! Solid angle (dimensionless quantity).
+
+quantity! {
+    /// Solid angle (dimensionless quantity).
+    quantity: SolidAngle; "solid angle";
+    /// Dimension of solid angle, 1 (dimensionless).
+    dimension: ISQ<
+        Z0,     // length
+        Z0,     // mass
+        Z0,     // time
+        Z0,     // electric current
+        Z0,     // thermodynamic temperature
+        Z0,     // amount of substance
+        Z0>;    // luminous intensity
+    kind: dyn (::si::marker::SolidAngleKind);
+    units {
+        /// SI derived unit of solid angle is steradians. It is the solid angle subtended at the
+        /// center of a unit sphere by a unit area on its surface.
+        @steradian: 1.0_E0; "sr", "steradian", "steradians";
+        @spat: 1.256_637_061_435_917_3_E1; "sp", "spat", "spats";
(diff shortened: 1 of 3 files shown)


src/si/solid_angle.rs
```diff
@@ -0,0 +1,108 @@
...
+            let r2: SolidAngle<V> = V::one().into();
+            let _: V = V::from(r1);
+            let _: V = r2.into();
+        }
+
+        #[test]
+        fn check_units() {
+            Test::assert_eq(&SolidAngle::new::<sa::steradian>(V::from_f64(4.0 * PI).unwrap()),
+                &SolidAngle::new::<sa::spat>(V::one()));
+            Test::assert_approx_eq(&SolidAngle::new::<sa::square_degree>(V::from_f64(360.0 * 360.0 / PI).unwrap()),
```
adamreichold, line 100: I was unable to make these exact, e.g. by moving π to the other side of the equation as `V::from_f64(PI).unwrap()`. Is there some numerical trick here I don't see?

iliekturtles:
Thanks for the PR! I'll do a code review today!

src/si/solid_angle.rs
iliekturtles: Floating point just doesn't have enough precision and the result ends up being 1 ulps (units in last place) off: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cd36961dfe4c57392d7421146e545951. We'll need to live with the approx equal assertions.

adamreichold:
@iliekturtles Applied the patch. Could you give me a hint on how you keep the formatting in check as rustfmt will not touch the stuff inside the macros reliably? (I have admittedly become dependent on rustfmt and hardly even consider formatting while writing code anymore.)

iliekturtles:
Build failed which is my fault! I was testing `assert_approx_eq` vs `assert_eq` and forgot to remove that before posting the patch. If you can switch the last three back to `assert_approx_eq` everything should work then I can merge for real.

For formatting I just spend a bunch of time looking at the code. I have vertical guidelines setup at the 100 character mark to see when text overflows and VIM's `gq` operator mostly works to re-wrap comments.

adamreichold:
Switched back to `assert_approx_eq`.

iliekturtles:
Thanks again for the PR. Merged!

PR #197, opened 2020-07-27
Remove `From`/`Into` impls for `Angle` and `SolidAngle`.

### LINKED ISSUE #188: Remove From/Into impls for Angle?
The impls were included as part of the original PR that added `Angle`. As @adamreichold brings up in #187 these are a potential source of errors as they assume that the value is always in radians.

@dunmatt Do you remember why the impls were added? Does your code use them?

https://github.com/iliekturtles/uom/blob/bdb73723da75a392aeb3f4d7afd1ce05af30886f/src/si/angle.rs#L95-L125

iliekturtles:
Impls were problematic because they assume the value is always in
radians. Using explicit conversions that include the unit should be used
instead. Impls were likely added when `Angle` was added and adapted from
`Ratio`. Resolves #188.

### DIFF
--- a/src/si/angle.rs
+++ b/src/si/angle.rs
@@ -129,38 +129,6 @@ where
     }
 }
 
-mod convert {
-    use super::*;
-
-    impl<U, V> ::lib::convert::From<V> for Angle<U, V>
-    where
-        U: ::si::Units<V> + ?Sized,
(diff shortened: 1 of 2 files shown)


adamreichold:
LGTM

iliekturtles:
Thanks! Merged.

PR #206, opened 2020-09-13
Convert `uom` to 2018 edition.

iliekturtles:
Generated code is still compatible with 2015 edition and the
`edition_check` test crate now verifies this.

If anyone is using `uom`'s macros to create new systems in their own 2015 crate I'd greatly appreciate testing against this branch. Everything should still be compatible.


iliekturtles:
Ran into a [Clippy](https://github.com/rust-lang/rust-clippy/issues/6096) issue and some vacation time. Hope to have this merged this week.

PR #209, opened 2020-10-06
Initial changes to create a `unit!` macro.

iliekturtles:
Additional changes still pending to allow the `quantity!` macro to also
be called outside the mod where the `system!` macro was called.

### DIFF
--- a/examples/unit.rs
+++ b/examples/unit.rs
@@ -1,0 +1,34 @@
+//! Example showing how to use the `unit!` macro to add new units to existing quantities.
+//!
+//! [Pull requests](https://github.com/iliekturtles/uom/pulls) for new units are always greatly
+//! appreciated.
+
+#[macro_use]
+extern crate uom;
+
+use uom::fmt::DisplayStyle::*;
+use uom::si::f32::*;
+use uom::si::length::meter;
+
+unit! {
+    system: uom::si;
+    quantity: uom::si::length;
+
+    @smoot: 1.702; "smoot", "smoot", "smoots";
+}
+
+fn main() {
+    let l1 = Length::new::<meter>(15.0);
+    let l2 = Length::new::<smoot>(1.0);
+
+    println!(
+        "{} = {}",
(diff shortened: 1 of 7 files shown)


examples/unit.rs
```diff
@@ -0,0 +1,26 @@
+//! Example showing how to use the `unit!` macro to add new units to existing quantities.
+
+#[macro_use]
+extern crate uom;
```
adamreichold, line 4: Maybe using the macro directly like
```
use uom::unit;
```
now that we are using edition 2018?

src/unit.rs
```diff
@@ -0,0 +1,257 @@
+/// Macro to implement a set of [measurement units][measurement]. Note that this macro must be
+/// executed in direct submodules of the module where the [`system!`](macro.system.html) macro was
+/// executed. `@...` match arms are considered private.
+///
+/// * `$unit`: Unit name (e.g. `meter`, `foot`).
+/// * `$conversion`: Conversion (coefficient and constant factor) from the unit to the base unit of
```
adamreichold, line 6: The wording of "constant factor" did puzzle me a bit during the first reading  "factor" suggests a product where a sum is computed. AFAIU, this is something like an offset or an intercept?

iliekturtles: Most units just use a coefficient. The constant factor is for thermodynamic temperature. Thoughts on better wording?

https://github.com/iliekturtles/uom/blob/1a3386a7f999c6da5a366366b094028848795ed2/src/si/thermodynamic_temperature.rs#L90-L93

iliekturtles:
Latest changes now ready for review.

I ended up stopping short of making the quantity! macro be able to be executed outside of direct submodules of the location where the system! macro was executed. Executing quantity! in an external crate runs into issues with attempting to implement external traits for external types and needing access to private items.

iliekturtles:
Note to self: Update the documentation to state that PRs for any new units are desired.

Issue #212, opened 2020-10-17
Supporting user-defined string parsing


bheisler:
uom's `Quantity::from_str` implementation is quite strict. It requires a number, a space and then an abbreviated unit. It does not support strings like "100g" or even "100 grams", even though these are natural ways for humans to write measurements.

I have implemented my own layer on top of uom's `from_str` to relax the requirement for a space, but there is no way for me to make it work for full units without manually listing all of the units in my code because uom doesn't seem to provide any functionality to work with full unit strings rather than abbreviations.

Could uom provide a function like `Mass::from_str_unit(100.0f64, "grams") -> Result<...>` (which would also accept "gram" and "g", but also "pounds" and "kg" and so on)? From my perspective, this would actually be better than changing `from_str`, as it makes it possible for me to implement my own human-friendly parsing code without needing to incorporate my opinions about human-friendly parsing into uom itself.

If this is not practical, could the `from_str` implementation be extended to allow the singular and plural units? Right now my human-friendly parsing layer works by creating a temporary string which does fit uom's desired format and then asking uom to parse that.

iliekturtles:
Adding descriptions to the current `FromStr` impl is probably a good start (PRs welcome). I would like to keep one parsing method that only accepts input based on SI conventions (see [this](https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-6-rules-and-style-conventions-printing-and-using) for one reference). A second could work with less precise input that would be more human friendly like you describe.

I took a brief look at `std::str` and it looks like `FromStr` is the only relevant trait in the standard library. Did I miss one? If making the split described above should `FromStr` be the strict or human parsing method?

bheisler:
FromStr is the only built-in parsing trait, yes. In my project I have my own second trait for providing more human-friendly parsing, but of course I don't expect uom to implement that; I'm just hoping for uom to expose some helpful functions that would make it easier for me to implement it myself.

I think Rust's documentation isn't entirely clear on how lenient or strict the FromStr implementation should be, so I think that decision should be left up to you. For my particular project, I need to define my own parsing regardless.

If I understand the code correctly, I think that all I'd need to do in order to add support for the singular and plural form of the unit is to add a few extra cases to the match in the macro that generates the quantities, is that correct? I can probably send a PR in the next few days.

iliekturtles:
I believe you're correct about just adding cases to the `FromStr` implementation. I took a brief look this morning and if you can submit a PR I'll review in more detail.

[...]

PR #214, opened 2020-10-20
Add singular and plural units to from_str impl. #212

bheisler:
This adds the singular and plural unit strings to the match in from_str, which means that strings containing them can be parsed.

I found that there are a number of quantities in uom which have at least one unit where the singular and plural is the same. There is also `pressure`, where pounds-per-square-inch is defined in two different ways. Because of this, the match sometimes includes multiple arms with the same string, resulting in a compiler warning. I've ignored this warning just on the generated match.

I've also updated the tests to check that the expected strings can be parsed.

### DIFF
--- a/src/quantity.rs
+++ b/src/quantity.rs
@@ -513,9 +513,7 @@ macro_rules! quantity {
 
                         #[allow(unreachable_patterns)]
                         match unit.trim() {
-                            $($abbreviation => Ok(Self::new::<super::super::$unit>(value)),)+
-                            $($singular => Ok(Self::new::<super::super::$unit>(value)),)+
-                            $($plural => Ok(Self::new::<super::super::$unit>(value)),)+
+                            $($abbreviation | $singular | $plural => Ok(Self::new::<super::super::$unit>(value)),)+
                             _ => Err(UnknownUnit),
                         }
                     }
--- a/src/quantity.rs
+++ b/src/quantity.rs
@@ -508,11 +508,14 @@ macro_rules! quantity {
                     fn from_str(s: &str) -> Result<Self, Self::Err> {
                         let mut parts = s.splitn(2, ' ');
                         let value = parts.next().unwrap();
-                        let abbr = parts.next().ok_or(NoSeparator)?;
+                        let unit = parts.next().ok_or(NoSeparator)?;
                         let value = value.parse::<V>().map_err(|_| ValueParseError)?;
 
-                        match abbr.trim() {
+                        #[allow(unreachable_patterns)]
+                        match unit.trim() {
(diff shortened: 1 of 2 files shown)


Aehmlo:
Seeing that it's an explicit test case, I gather that e.g. `"1 meters"` parses. I just want to make sure that this is the desired end result. In my applications, I would expect an attempt to parse such a string to be suggestive of a bug elsewhere. I do suspect that checking this edge case may be nontrivial, though, and we probably don't want to get too far into input validation. I guess it comes down to the concept of a balance between human-friendly, permissive parsing vs. stricter parsing discussed in #212.

bheisler:
I think that disallowing "1 meters" would require a strange edge-case in the code for little to no benefit, so I would recommend allowing it. It's up to you though.

Validation rules like that perhaps make more sense to define at the application level.

src/quantity.rs
```diff
@@ -508,11 +508,14 @@ macro_rules! quantity {
...
                         let value = value.parse::<V>().map_err(|_| ValueParseError)?;
 
-                        match abbr.trim() {
+                        #[allow(unreachable_patterns)]
+                        match unit.trim() {
                             $($abbreviation => Ok(Self::new::<super::super::$unit>(value)),)+
+                            $($singular => Ok(Self::new::<super::super::$unit>(value)),)+
+                            $($plural => Ok(Self::new::<super::super::$unit>(value)),)+
                             _ => Err(UnknownUnit),
                         }
```
iliekturtles, line 520: The current code can be simplified:

```rust
                        match unit.trim() {
                            $($abbreviation | $singular | $plural => Ok(Self::new::<super::super::$unit>(value)),)+
                            _ => Err(UnknownUnit),
                        }
```

We could also consider only matching singular when `value.is_one()` and plural when not. This would require separate matches as written currently in order to fit the match guards. If we do end up going this route then only one macro iteration `$(...)+` is needed around all three lines instead of one on each line.

iliekturtles (commented):
Thanks for the PR! I'll review in more depth tomorrow. I feel like there should be more test cases but I'm not sure what they could be yet.

src/quantity.rs
adamreichold: > We could also consider only matching singular when value.is_one() and plural when not. 

As @bheisler said above, being strict in this sense seems to provide little benefit especially when parsing user input. I would suggest erring on the side of lenience here following the parsing adage "be lenient in what you accept, be strict in what you generate".

bheisler: I've updated it to simplify the code, as suggested.

iliekturtles:
Thanks again for the PR!

PR #215, opened 2020-10-21
Generate an enum with one variant for each unit, use that for iteration.

bheisler:
This extends the `quantity!` macro to generate a non-exhaustive enum with one variant per unit. A static slice of the different potential values of this enum is also generated, which is exposed publicly through a `units()` function, which returns an iterator. This can be used to iterate over the different units.

### DIFF
--- a/src/quantity.rs
+++ b/src/quantity.rs
@@ -288,43 +288,41 @@ macro_rules! quantity {
         #[allow(non_camel_case_types)]
         #[non_exhaustive]
         #[derive(Debug, Clone, Copy)]
-        pub enum UnitEnum {
+        pub enum Units {
             $(#[doc=$plural] $unit($unit),)+
         }
-        impl UnitEnum {
+
+        impl Units {
             /// Unit abbreviation.
-            #[inline]
             pub fn abbreviation(&self) -> &'static str {
                 match self {
-                    $(UnitEnum::$unit(_) => <$unit as super::Unit>::abbreviation(),)+
+                    $(Units::$unit(_) => <$unit as super::Unit>::abbreviation(),)+
                 }
             }
 
             /// Unit singular description.
-            #[inline]
             pub fn singular(&self) -> &'static str {
                 match self {
-                    $(UnitEnum::$unit(_) => <$unit as super::Unit>::singular(),)+
+                    $(Units::$unit(_) => <$unit as super::Unit>::singular(),)+
                 }
             }
 
             /// Unit plural description.
-            #[inline]
             pub fn plural(&self) -> &'static str {
                 match self {
-                    $(UnitEnum::$unit(_) => <$unit as super::Unit>::plural(),)+
+                    $(Units::$unit(_) => <$unit as super::Unit>::plural(),)+
                 }
             }
         }
 
-        static ALL_UNIT_ENUMS: &[UnitEnum] = &[
-            $(UnitEnum::$unit($unit),)+
+        static ALL_UNITS: &[Units] = &[
+            $(Units::$unit($unit),)+
         ];
 
         /// Iterate over all defined units for this quantity.
         #[inline]
-        pub fn units() -> impl Iterator<Item=UnitEnum> {
-            ALL_UNIT_ENUMS.iter().copied()
+        pub fn units() -> impl Iterator<Item = Units> {
+            ALL_UNITS.iter().copied()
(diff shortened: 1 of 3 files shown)


adamreichold:
This indeed much simpler than doing tricks with `HList`! I think a test case that exercises the API would be nice.

src/quantity.rs
```diff
@@ -284,6 +284,49 @@ macro_rules! quantity {
...
+        /// Unit enum.
+        #[allow(non_camel_case_types)]
+        #[non_exhaustive]
+        #[derive(Debug, Clone, Copy)]
+        pub enum UnitEnum {
+            $(#[doc=$plural] $unit($unit),)+
+        }
+        impl UnitEnum {
+            /// Unit abbreviation.
+            #[inline]
```
iliekturtles, line 296:
```suggestion
```

I don't think we want these methods marked `inline`.

src/quantity.rs
```diff
@@ -284,6 +284,49 @@ macro_rules! quantity {
             $description
         }
 
+        /// Unit enum.
+        #[allow(non_camel_case_types)]
+        #[non_exhaustive]
+        #[derive(Debug, Clone, Copy)]
+        pub enum UnitEnum {
```
iliekturtles, line 291:
```suggestion
        pub enum Units {
```

Seems redundant to include `Enum` in the name. Does `Units` work without colliding with any other types?

src/quantity.rs
```diff
@@ -284,6 +284,49 @@ macro_rules! quantity {
...
+            /// Unit plural description.
+            #[inline]
+            pub fn plural(&self) -> &'static str {
+                match self {
+                    $(UnitEnum::$unit(_) => <$unit as super::Unit>::plural(),)+
+                }
+            }
+        }
+
+        static ALL_UNIT_ENUMS: &[UnitEnum] = &[
```
iliekturtles, line 320:
```suggestion
        static ALL_UNITS: &[UnitEnum] = &[
```

`ENUMS` redundant?

src/quantity.rs
```diff
@@ -284,6 +284,49 @@ macro_rules! quantity {
...
+            $(UnitEnum::$unit($unit),)+
+        ];
+
+        /// Iterate over all defined units for this quantity.
+        #[inline]
+        pub fn units() -> impl Iterator<Item=UnitEnum> {
```
iliekturtles, line 326:
```suggestion
        pub fn units() -> impl Iterator<Item = UnitEnum> {
```

iliekturtles (commented):
Thanks for the PR!

bheisler:
I think I've fixed everything.

src/quantity.rs
```diff
@@ -284,6 +284,47 @@ macro_rules! quantity {
...
+                }
+            }
+        }
+
+        static ALL_UNITS: &[Units] = &[
+            $(Units::$unit($unit),)+
+        ];
+
+        /// Iterate over all defined units for this quantity.
+        #[inline]
```
iliekturtles, line 323:
```suggestion
```

I don't think `inline` is needed here either.

src/si/volume.rs
```diff
@@ -105,6 +105,19 @@ quantity! {
 
 #[cfg(test)]
 mod tests {
+    #[test]
+    fn test_volumn_units() {
```
iliekturtles, line 109:
```suggestion
    fn test_volume_units() {
```

Because this test is really for the `Units` enum and `units()` method move to `src/tests/quantity.rs`. I also suggest renaming to something like `test_units_enum` or `test_units_iterator`.

iliekturtles (commented):
Couple more minor changes. I think the code looks good so far but I have only reviewed through github. I want to do one more review, hopefully this weekend, with the compiler then I think we'll be able to merge.

iliekturtles:
I haven't forgotten about this, I just didn't finish my review this weekend. I like using `$plural` for the doc comment and think it would be good to switch `struct $unit` to use the same value if there is no user-provided comment and use the user-provided comment on the enum.

iliekturtles:
Merged! Thanks so much for the PR and sorry for the delays. My quest to use the user-provided documentation on the `enum` members ended up being a dead-end because you can't use a macro in place of an enum member. I will submit a separate PR to convert `struct $unit` to use `$plural`.

PR #228, opened 2020-12-09
Ensure `UnwindSafe` and `RefUnwindSafe` are implemented.

iliekturtles:
This fix also simplifies ensuring that `Unpin` is implemented by using
trait inheritance instead of explicit impls.

@DusterTheFirst, if you could test with your code I would great appreciate it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -246,6 +246,15 @@ pub mod lib {
         pub use std::ops::*;
         pub use typenum::type_operators::*;
     }
+
+    // Export `panic` module when the `std` feature is not enabled. `RefUnwindSafe` and `UnwindSafe`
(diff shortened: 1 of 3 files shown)


DusterTheFirst:
It seems to be working on my code!

iliekturtles:
Excellent, thanks for testing! I'll give it a day or so before merging in case anyone else has a chance to review. After that I think I'm going to try to get a new release out to crates.io in the next week or so.

iliekturtles:
Ended up running into an issue with `no_std` since `UnwindSafe` is `std` only.

https://www.reddit.com/r/rust/comments/k852ac/hey_rustaceans_got_an_easy_question_ask_here/gf7273z/

Issue #201, opened 2020-08-07
Support for logarithmic units


Atmelfan:
While not strictly supported by SI it's common in electronics and other industries to use logarithmic quantities expressed in Decibels such as `dBm[W]`, `dBuV/m`.

It would be nice to have a decibel_<unit> (Ex: decibel_milliwatt) for these common usages which converts them to the linear base quantities or vice versa.

It would need to perform different conversion to power and field quantities (see below).

Power quantities (watt):
```
P = 10.powf(Lp/10)*P0
Lp = 10*log(P/P0)
```

Field quantities (volt, ampere etc):
```
F = 10.powf(Lf/20)*F0
Lf = 20*log(F/F0)
```

Where:
 * F/P = Linear quantity
 * Lf/Lp = Logarithmic quantity in decibel
 * F0/P0 = Reference quantity. 

Usually the reference is provided as a suffix behind dB, e.g. dBuV => F0 = 1uV.

The simple decibel (dB) and derivatives (dBi etc) would also be applicable as a Ratio unit.

The Bell (B) and Neper (Np) could also be supported but nobody cares about those.

Not sure how hard this would be to implement as it would not be a simple coefficient and offset.

A reasonable alternative would be a special functions like `new_db::<reference>(logvalue)`  and `get_db::<reference>()` but would not be as versatile and consistent.

If I can understand how all the macros and generics works I'll try to implement something...

iliekturtles:
I've only briefly looked at the problem so far. I think what would need to be done is to pull conversion formulas (`from_base`, `to_base`, `change_base`) into `Conversion<V>` (or sub-traits). The default could still do a linear conversion, but certain units could override the default calculation with more complex formulas.

iliekturtles:
https://github.com/iliekturtles/uom/pull/229#issuecomment-1169055412

> Logarithmic units require `pow`/`root` functions which are not available in no-std/no-libm builds. Excluding logarthmic units has proven difficult (in-place filter foiled by macros unable to be used inside an enum definition). Using a tt-muncher could possibly solve the issue, but would likely hit the recursion limit very quickly and likely the quadratic performance characteristics would make `uom`'s slow compilation even slower.

I just haven't had a chance to get back to this since then.

sww1235:
This could also be used to implement AWG for area, as it has a logarithmic relationship with mm^2 .

PR #229, opened 2020-12-12
Added logarithmic units

Atmelfan:
Poor attempt at implementing #201.

No tests, doesn't compile under no_std, missing documentation, fails CI, etc. Breaks every rule regarding macros and generic programming. 

Probably as far as my brain can take this.
image

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -396,7 +396,8 @@ pub enum ConstantOp {
 ///
 /// [units]: http://jcgm.bipm.org/vim/en/1.13.html
 /// [factor]: https://jcgm.bipm.org/vim/en/1.24.html
-pub trait Conversion<V> {
+pub trait Conversion<V> where
+    V: Conversion<V, T = Self::T> {
     /// Conversion factor type specific to the underlying storage type.
     type T: ConversionFactor<V>;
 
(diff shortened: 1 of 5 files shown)


iliekturtles:
Thanks for the PR! My plan right now is to finish up the two other open PRs thenpush out a new version. After that I'll dig into this PR in more detail.

iliekturtles:
I finally released v0.31.0 earlier today and will start reviewing this now. Took me a bit longer than expected!

iliekturtles:
Could you possibly rebase this on top of master? The recent [change](https://github.com/iliekturtles/uom/commit/0e597dd21d527d43848750d3e8467d676cc1ae2c) to extract some of the code from quantity.rs into unit.rs is causing conflicts with this PR. I started looking to do this but never ended up finishing with the limited time I've had recently.

Atmelfan:
I'll take a look over the weekend.

iliekturtles:
Finally got a chance to rebase. https://github.com/iliekturtles/uom/compare/Atmelfan-master Will start reviewing for real now!

iliekturtles:
I finally finished my library book, which I'm going to blame any delays on. I have been reading about power and root-power quantities and digging in to this more and wanted to provide an update. The trait changes mostly need minor fixes like documentation and formating. I believe the `to`/`from_linear` functions should be merged into `to`/`from`/`change_base`. I'm reviewing rustc/llvm optimizations next to ensure that doing so will keep these functions zero-cost.

iliekturtles:
@Atmelfan, I'm starting to look at integrating these changes again after finally merging #260 and was wondering why the `Ratio`'s `decibel` is defined as a root-power ratio (`10^(1/10)`) but `Power`'s `decibel-watt` derivatives use power ratios (`10^(1/20)`).

iliekturtles:
Logarithmic units require `pow`/`root` functions which are not available
in no-std/no-libm builds. Excluding logarthmic units has proven
difficult (in-place filter foiled by macros unable to be used inside an
enum definition). Using a tt-muncher could possibly solve the issue, but
would likely hit the recursion limit very quickly and likely the
quadratic performance characteristics would make `uom`'s slow
compilation even slower.

dzmitry-lahoda:
so need to merge libm support first https://github.com/iliekturtles/uom/pull/530 ? 

i think compilation speed and macro ide support become way better during last 3 years.

PR #250, opened 2021-04-10
RFC: Add a zero constant to quantities based on primitive types.

### LINKED ISSUE #26: Add zero constants/methods for each quantity
Add `constant`? Add `const fn`? `impl Zero`?

adamreichold:
The public helper trait is a bit cumbersome but it appears to be the most straight-forward way to construct a zero value in const context.

An alternative that avoid the intermediate trait this would be to implement something like
```rust
impl<U> $quantity<U, $type>
where
    U: __system::Units<$type> + ?Sized,
{
    /// Constant representing the zero value.
    pub const ZERO: Self = Self {
        value: 0 as $type,
    };
}
```
for the same set of primitive types.

Fixes #26

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -456,6 +456,20 @@ pub trait ConversionFactor<V>:
     fn value(self) -> V;
 }
 
+/// Helper trait to identify the zero value of a type at compile time.
+///
+#[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
+#[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
+/// # use uom::si::f32::Length;
+/// use uom::ConstZero;
+///
+/// const ORIGIN: (Length, Length, Length) = (Length::ZERO, Length::ZERO, Length::ZERO);
+/// ```
+pub trait ConstZero {
+    /// Constant representing the zero value.
+    const ZERO: Self;
+}
+
 /// Default [kind][kind] of quantities to allow addition, subtraction, multiplication, division,
 /// remainder, negation, and saturating addition/subtraction.
 ///
@@ -507,6 +521,10 @@ storage_types! {
             self
         }
     }
+
+    impl crate::ConstZero for V {
+        const ZERO: Self = 0.0;
+    }
 }
 
 storage_types! {
@@ -531,6 +549,10 @@ storage_types! {
         fn value(self) -> V {
             self.to_integer()
         }
+    }
+
+    impl crate::ConstZero for V {
+        const ZERO: Self = 0;
     }
 }
 
--- a/src/system.rs
+++ b/src/system.rs
@@ -1288,6 +1288,19 @@ macro_rules! system {
             }
         }
 
+        impl<D, U, V> $crate::ConstZero for Quantity<D, U, V>
+        where
+            D: Dimension + ?Sized,
+            U: Units<V> + ?Sized,
+            V: $crate::num::Num + $crate::Conversion<V> + $crate::ConstZero,
+        {
+            const ZERO: Self = Self {
+                dimension: $crate::lib::marker::PhantomData,
+                units: $crate::lib::marker::PhantomData,
+                value: V::ZERO,
+            };
+        }
+
         serde! {
         impl<D, U, V> $crate::serde::Serialize for Quantity<D, U, V>
         where


src/storage_types.rs
```diff
@@ -248,3 +248,18 @@ storage_type_types! {
...
+        impl $crate::ConstZero for $type {
+            #[allow(trivial_numeric_casts)]
+            const ZERO: Self = 0 as Self;
+        }
+
+        impl_const_zero!($($rest)*);
+    };
+}
+
+impl_const_zero!(usize u8 u16 u32 u128 isize i8 i16 i32 i64 i128 f32 f64);
```
iliekturtles, line 265: Any reason not to use `storage_types!`?

```suggestion

storage_types! {
    types: PrimInt, Float;

    impl $crate::ConstZero for V {
        #[allow(trivial_numeric_casts)]
        const ZERO: Self = 0 as Self;
    }
}
```

src/quantity.rs
```diff
@@ -391,6 +391,19 @@ macro_rules! quantity {
...
+            U: __system::Units<V> + ?Sized,
+            V: $crate::num::Num + $crate::Conversion<V> + $crate::ConstZero,
+        {
+            /// Constant representing the zero value.
+            pub const ZERO: Self = Self {
+                dimension: $crate::lib::marker::PhantomData,
+                units: $crate::lib::marker::PhantomData,
+                value: V::ZERO,
+            };
+        }
```
iliekturtles, line 405: Could this impl block be merged with one in `system.rs` and the `const` changed to a `const fn`? Would both optimize down to the same inlined code anyway?

```rust
const fn zero() -> Self
    where V: $crate::ConstZero
{
    ....
}
```

src/storage_types.rs
adamreichold: None at all. I have to admit to not understanding the storage types macro sufficiently to produce the above suggestion on my own.

src/quantity.rs
adamreichold: > Could this impl block be merged with one in system.rs and the const changed to a const fn?

I will try, but I fear that going for a `const fn` will bring back the unstable trait bounds on `const fn` problem described in https://github.com/iliekturtles/uom/issues/26#issuecomment-614718476

> Would both optimize down to the same inlined code anyway?

That should definitely be the case as all evaluation happens at compile time with the resulting bits being "burned into" the executable.

adamreichold: Indeed, a `const fn` seems out of reach on stable Rust for now. But I can move the impl block to `system.rs` next to the impl block for the `Zero` trait and actually turn it into an impl block for the `ConstZero` trait so that symmetry is preserved.

iliekturtles (commented):
Thanks so much for the PR! I didn't intend to do a full review yesterday morning and completely forgot to thank you.

src/storage_types.rs
```diff
@@ -248,3 +248,12 @@ storage_type_types! {
...
 }
+
+storage_types! {
+    types: PrimInt, Float;
+
+    impl crate::ConstZero for V {
+        #[allow(trivial_numeric_casts)]
+        const ZERO: Self = 0 as Self;
+    }
+}
```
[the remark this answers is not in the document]
adamreichold: Changed as suggested.

src/lib.rs
```diff
@@ -456,6 +456,12 @@ pub trait ConversionFactor<V>:
     fn value(self) -> V;
 }
 
+/// Helper trait to identify the zero value of a type at compile time.
+pub trait ConstZero {
```
iliekturtles, line 460: We should have at least a simple doc example. Perhaps it can be improved in some way?

```suggestion
/// Helper trait to identify the zero value of a type at compile time.
///
#[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
#[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
/// # use uom::si::f32::Length;
/// use uom::ConstZero;
/// 
/// let zero = Length::ZERO;
/// ```
pub trait ConstZero {
```

iliekturtles (commented):
Couple of minor things and I think we're good to merge.

Thoughts about the proposed `new_in_base_units`. Just push that into the const fn issue or create a new issue?

adamreichold:
> Thoughts about the proposed new_in_base_units. Just push that into the const fn issue or create a new issue?

I think a new issue is the better approach because our goal should probably be to have `Quantity::new` be a const fn even though this is currently still blocked on the trait bounds as well as const eval for floating point math. We could use a new issue to e.g. discuss adding nightly only features or macro-based workarounds.

src/lib.rs
adamreichold: Committed a slightly extended version that constructs a new constant using `ZERO` constants.

iliekturtles:
Thanks so much for this PR. I'll create a new issue to continue efforts for `new` / `new_from_base_units`

PR #258, opened 2021-06-11
Deny `rustc` warnings during full test suite and min test matrix.

iliekturtles:
`RUSTFLAGS="-D warnings"` is set in both workflows to ensure that
`rustc` warnings and caught and fail the respective workflows.


iliekturtles:
Only failures are the pre-existing `memchr` issue. Merging.

PR #263, opened 2021-08-04
Molar heat capacity.

iliekturtles:
Initial work to setup the quantity. Still need to add additional units and tests.

@nakedible Any specific units you're looking for? If you could review I would greatly appreciate it.


iliekturtles:
Build failures are limited to 1.37 (to be fixed in another PR) and there are no other unit suggestions so I'm going to merge!

Issue #163, opened 2019-08-26
Parse units from a string representation


stephanbuys:
Hi, is there a way to parse string representations into the appropriate measurements? Ie, representing "12lbs" or "3kg" as mass?

Perhaps a companion crate to do so?

stephanbuys:
Thanks!

PR #277, opened 2021-12-17
Allow parsing quantities without a separator between value and unit

mox-mox:
## Summary
This PR allows uom to read quantities from a string even if it does not contain a space between value and unit.

## Rationale
I was trying to read some SI units from the command line.  The current implementation of `Quantity::from_str` requires a space between value and unit, e.g. `3 kg` will parse properly but `3kg` will not.
But command lines tend to break arguments at whitespace, so reading in `3 kg` would mean I'd have to read in two arguments ("3" and "kg") and then merge them to let uom parse them. Or I could force every user of my program to quote the input, e.g. `$ ./my_prog --weight "3 kg"` instead of `./my_prog --weight 3 kg`.

Or I could accept `./my_prog --weight 3kg`. I believe this could be helpful for more than just my situation, as evidenced by previous issues like #163 or #212. So, I've made this pull request.

## Changes
 - Edit the implementation of `FromStr` for `quantity!` to cope with strings with no space as a fall-back.
     + If there is a space, the algorithm proceeds as before.
     + If there isn't, the quantity is expressed as only one word. The algorithm tries to remove each unit from the string.
     + If this succeeds and the remaining string can be parsed as a number, we have retrieved the correct unit and value, so the algorithm can return the appropriate quantity as if there was a space to begin with.
 - Remove the `NoSeparator` variant from `ParseQuantityError` because is does not have any purpose anymore if the omitted space is not an error anymore.
 - Edit the unit tests to expect everything that worked with a space before to also work without a space.

p.s.> Nice and very handy crate. Thank you for all the work that went into it!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -658,9 +658,6 @@ pub mod str {
     #[allow(missing_copy_implementations)]
     #[derive(Clone, Debug, Eq, PartialEq)]
     pub enum ParseQuantityError {
-        /// No separators (spaces) were encountered.
-        NoSeparator,
-
         /// An error occurred while parsing the value (first) portion of the string.
         ///
         /// Due to exhaustiveness and type system limitations, this variant does not encode
@@ -680,7 +677,6 @@ pub mod str {
             use ParseQuantityError::*;
 
             match *self {
-                NoSeparator => write!(f, "no space between quantity and units"),
                 ValueParseError => write!(f, "error parsing unit quantity"),
                 UnknownUnit => write!(f, "unrecognized unit of measure"),
             }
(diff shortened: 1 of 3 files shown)


mox-mox:
If preferred, I could also hide this behind a new feature like `parse_no_space`.

iliekturtles:
Thanks for the PR, I'll review this weekend.

mox-mox:
Hey, I hope you had a nice holiday season and new years eve :)
Can I help the review somehow? Do you need some more information?

iliekturtles:
Sorry for taking so long to get back to this. My current thoughts are to keep the strict rules defined by [NIST](https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-6-rules-and-style-conventions-printing-and-using) in the `FromStr` impl and delegate alternate rules like this PR implements to some kind of alternate parsing method. I just opened #278 based on past discussions when parsing was previously discussed.

Thoughts?

iliekturtles:
Thank you again for the PR, however I'm going to close without merging. I want to keep the strict rules defined by NIST in the `FromStr` impl. A new PR based on the ideas in #278 is always welcome.

PR #285, opened 2022-02-28
fix: use singular unit names in si::absement

nick-pascucci-spire:
I just noticed that the naming convention was inconsistent as I was starting to use the new module. This should clear it up!


iliekturtles:
Good catch! Thanks for the PR .

Issue #286, opened 2022-03-10
Can I use complex numbers (num-complex) with uom?


TobTobXX:
Now, I know this is unusual, but it actually comes up in quantum physics. I tried some stuff like using `uom::si::quantities::Length<Complex64>`, but then the compiler tells me that `uom::Conversion` somehow isn't implemented.

Am I missing something? Or is this indeed something that could be worked on in the uom library?

iliekturtles:
`uom` can't currently support arbitrary underlying storage types such as `Complex64`. See #122 for context. Since [`Complex<T>`](https://docs.rs/num/0.4.0/num/struct.Complex.html) is part of `num` adding support for `Complex32` and `Complex64` shouldn't be too much effort.

For anyone interested in submitting a PR see e2d5f20e4f55ec06bcfaee7299e5f18c33617706 for the last time an underlying storage type was added. I expect `Complex` implementations would look like a cross between `f32`/`f64` and `Rational32`/`Rational64`.

TobTobXX:
Fixed by #287.

Issue #44, opened 2017-12-24
test failures with integers enabled


radix:
I didn't realize that I didn't have integers enabled when running tests until I tried to write integer-specific tests for the new Saturating support I'm working on. When I tried to run them, I got these failures (with master):

```
    si::acceleration::tests::u32::check_units
    si::area::tests::u32::check_units
    si::force::tests::u32::check_units
    si::frequency::tests::u32::check_units
    si::velocity::test::u32::check_units
    si::volume::tests::u32::check_units
    tests::system_macro::op_assign::u32::sub_assign
    tests::system_macro::u32::sub
```

It seems that travis isn't running tests with ints enabled either.

iliekturtles:
Known issue briefly mentioned in the v0.16.0 release notes. All the `check_units` failures are because conversion factors exceed the underlying storage type limits. Others like the `u32::sub_assign` and `u32::sub` failures are because quickcheck is generating test values that have a signed result once the operator is applied.

My plan is to implement `PartialEq`, `Eq`, `PartialOrd`, and `Ord` and then figure out how to resolve these issues.

PR #287, opened 2022-03-10
[WIP] Add support for Complex number storage types.

TobTobXX:
I have no idea what I'm doing.

Based on e2d5f20e4f55ec06bcfaee7299e5f18c33617706. Addresses #286.

Todo:
 - [X] Code compiles
 - [x] Code works (just some light testing in a simple project)
 - [x] Tests

Discuss before merge:
 - [x] Also support `Complex<T>`, or maybe only `Complex<T>` instead of `Complex32`, `Complex64`?
 - [x] Are the features correct like this? I have never used features /shrug

### DIFF
--- a/.github/workflows/ci-full-test-suite.yml
+++ b/.github/workflows/ci-full-test-suite.yml
@@ -54,10 +54,10 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --verbose --no-run --no-default-features --features "autoconvert usize isize bigint bigrational si std use_serde"
+          args: --verbose --no-run --no-default-features --features "autoconvert usize isize bigint bigrational complex32 si std use_serde"
 
       - name: Test all non-si features
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --verbose --no-run --no-default-features --features "autoconvert usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 bigint biguint rational rational32 rational64 bigrational f32 f64 std use_serde"
+          args: --verbose --no-run --no-default-features --features "autoconvert usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 bigint biguint rational rational32 rational64 bigrational complex32 complex64 f32 f64 std use_serde"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ members = [
 num-traits = { version = "0.2", default-features = false }
 num-rational = { version = "0.4", optional = true, default-features = false }
 num-bigint = { version = "0.4", optional = true, default-features = false, features = ["std"] }
+num-complex = { version = "0.4", optional = true, default-features = false, features = ["std"] }
 serde = { version = "1.0", optional = true, default-features = false }
 typenum = "1.13"
 
@@ -66,6 +67,8 @@ rational = ["rational-support"]
 rational32 = ["rational-support"]
 rational64 = ["rational-support"]
 bigrational = ["bigint-support"]
+complex32 = ["complex-support"]
+complex64 = ["complex-support"]
 f32 = []
 f64 = []
 si = []
@@ -81,6 +84,7 @@ use_serde = ["serde"]
 # Internal features to include appropriate num-* crates.
 rational-support = ["num-rational"]
 bigint-support = ["num-bigint", "num-rational/num-bigint-std"]
+complex-support = ["num-complex"]
 
 [[example]]
 name = "base"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,6 +74,7 @@
 //!         "isize", "i8", "i16", "i32", "i64", "i128", # Signed integer storage types.
 //!         "bigint", "biguint", # Arbitrary width integer storage types.
 //!         "rational", "rational32", "rational64", "bigrational", # Integer ratio storage types.
+//!         "complex32", "complex64", # Complex floating point storage types.
 //!         "f32", "f64", # Floating point storage types.
 //!         "si", "std", # Built-in SI system and std library support.
 //!         "use_serde", # Serde support.
@@ -86,11 +87,10 @@
 //!    directly interact. The feature exists to account for compiler limitations where zero-cost
 //!    code is not generated for non-floating point underlying storage types.
 //!  * `usize`, `u8`, `u16`, `u32`, `u64`, `u128`, `isize`, `i8`, `i16`, `i32`, `i64`, `i128`,
-//!    `bigint`, `biguint`, `rational`, `rational32`, `rational64`, `bigrational`, `f32`, `f64` --
-//!    Features to enable underlying storage types. At least one of these features must be enabled.
-//!    `f32` and `f64` are enabled by default. See the [Design](#design) section for implications
-//!    of choosing different
-//!    underlying storage types.
+//!    `bigint`, `biguint`, `rational`, `rational32`, `rational64`, `bigrational`, `complex32`,
+//!    `complex64`, `f32`, `f64` -- Features to enable underlying storage types. At least one of
+//!    these features must be enabled. `f32` and `f64` are enabled by default. See the
+//!    [Design](#design) section for implications of choosing different underlying storage types.
 //!  * `si` -- Feature to include the pre-built [International System of Units][si] (SI). Enabled by
 //!    default.
 //!  * `std` -- Feature to compile with standard library support. Disabling this feature compiles
@@ -122,7 +122,8 @@
 //! quantity. Alternative base units can be used by executing the macro defined for the system of
 //! quantities (`ISQ!` for the SI). `uom` supports `usize`, `u8`, `u16`, `u32`, `u64`, `u128`,
 //! `isize`, `i8`, `i16`, `i32`, `i64`, `i128`, `bigint`, `biguint`, `rational`, `rational32`,
-//! `rational64`, `bigrational`, `f32`, and `f64` as the underlying storage type.
+//! `rational64`, `bigrational`, `complex32`, `complex64`, `f32`, and `f64` as the underlying
+//! storage type.
 //!
 //! A consequence of normalizing values to the base unit is that some values may not be able to be
 //! represented or can't be precisely represented for floating point and rational underlying
@@ -199,6 +200,7 @@
     feature = "i128",
     feature = "bigint", feature = "biguint",
     feature = "rational", feature = "rational32", feature = "rational64", feature = "bigrational",
+    feature = "complex32", feature = "complex64",
     feature = "f32", feature = "f64", )))]
 compile_error!("A least one underlying storage type must be enabled. See the features section of \
     uom documentation for available underlying storage type options.");
@@ -215,13 +217,20 @@ pub extern crate num_bigint;
 pub extern crate num_rational;
 
 #[doc(hidden)]
+#[cfg(feature = "complex-support")]
+pub extern crate num_complex;
+
+#[doc(hidden)]
 #[cfg(feature = "serde")]
 pub extern crate serde;
 
 #[doc(hidden)]
 pub extern crate typenum;
 
-#[cfg(all(test, any(feature = "f32", feature = "f64")))]
+#[cfg(all(
+    test,
+    any(feature = "f32", feature = "f64", feature = "complex32", feature = "complex64")
+))]
 #[macro_use]
 extern crate approx;
 #[cfg(test)]
@@ -283,6 +292,11 @@ pub mod num {
     #[cfg(any(feature = "rational-support", feature = "bigint-support"))]
     pub mod rational {
         pub use num_rational::*;
+    }
+
+    #[cfg(feature = "complex-support")]
+    pub mod complex {
+        pub use num_complex::*;
     }
 }
 
@@ -641,6 +655,42 @@ storage_types! {
         #[inline(always)]
         fn value(self) -> V {
             self
+        }
+    }
+}
+
+storage_types! {
+    types: Complex;
+    impl crate::Conversion<V> for V {
+        type T = VV;
+
+        #[inline(always)]
+        fn constant(op: crate::ConstantOp) -> Self::T {
+            match op {
+                crate::ConstantOp::Add => -<Self::T as crate::num::Zero>::zero(),
+                crate::ConstantOp::Sub => <Self::T as crate::num::Zero>::zero(),
+            }
+        }
+
+        #[inline(always)]
+        fn conversion(&self) -> Self::T {
+            // Conversion factor is the norm of the number. Scaling with length again yields the
+            // same number.
+            self.norm()
+        }
+    }
+
+    impl crate::ConversionFactor<V> for VV {
+        #[inline(always)]
+        fn powi(self, e: i32) -> Self {
+            self.powi(e)
+        }
+
+        #[inline(always)]
+        fn value(self) -> V {
+            // Conversion by scaling (multiplication with only real number). Scaling a normalized
+            // number yields the original number again.
+            V::new(self, 0.0)
         }
     }
 }
--- a/src/storage_types.rs
+++ b/src/storage_types.rs
@@ -7,8 +7,8 @@
 /// * `$T`: Types to generate a module for. Accepts all underlying storage types along with a number
 ///   of different categories:
 ///   * `All`: `usize`, `u8`, `u16`, `u32`, `u64`, `u128`, `isize`, `i8`, `i16`, `i32`, `i64`,
-///     `i128`, `BigInt`, `BigUint`, `Rational`, `Rational32`, `Rational64`, `BigRational`, `f32`,
-///     and `f64`.
+///     `i128`, `BigInt`, `BigUint`, `Rational`, `Rational32`, `Rational64`, `BigRational`,
+///     `Complex32`, `Complex64`, `f32`, and `f64`.
 ///   * `PrimInt`: `usize`, `u8`, `u16`, `u32`, `u64`, `u128`, `isize`, `i8`, `i16`, `i32`, `i64`,
 ///     and `i128`.
 ///   * `Ratio`: `Rational`, `Rational32`, `Rational64`, and `BigRational`.
@@ -16,6 +16,7 @@
 ///   * `Signed`: `isize`, `i8`, `i16`, `i32`, `i64`, `i128`, `BigInt`, `Rational`, `Rational32`,
 ///     `Rational64`, `BigRational`, `f32`, and `f64`.
 ///   * `Unsigned`: `usize`, `u8`, `u16`, `u32`, `u64`, `u128`, and `BigUint`.
+///   * `Complex`: `Complex32` and `Complex64`.
 /// * `$tt`: Code to place into each storage type module.
 ///
 #[cfg_attr(all(feature = "f32", feature = "f64"), doc = " ```rust")]
@@ -100,6 +101,20 @@ macro_rules! storage_types {
     (@type ($(#[$attr:meta])*) @$M:ident BigRational ($($tt:tt)*)) => {
         storage_type_bigrational!(($(#[$attr])*) @$M ($($tt)*));
     };
+    (@type ($(#[$attr:meta])*) @$M:ident Complex32 ($($tt:tt)*)) => {
+        storage_type_complex32!(($(#[$attr])*) @$M (
+            /// Inner storage type.
+            #[allow(dead_code)]
+            pub type VV = f32;
+            $($tt)*));
+    };
+    (@type ($(#[$attr:meta])*) @$M:ident Complex64 ($($tt:tt)*)) => {
+        storage_type_complex64!(($(#[$attr])*) @$M (
+            /// Inner storage type.
+            #[allow(dead_code)]
+            pub type VV = f64;
+            $($tt)*));
+    };
     (@type ($(#[$attr:meta])*) @$M:ident f32 ($($tt:tt)*)) => {
         storage_type_f32!(($(#[$attr])*) @$M ($($tt)*));
     };
@@ -125,6 +140,8 @@ macro_rules! storage_types {
         storage_types!(@type ($(#[$attr])*) @$M Rational32 ($($tt)*));
         storage_types!(@type ($(#[$attr])*) @$M Rational64 ($($tt)*));
         storage_types!(@type ($(#[$attr])*) @$M BigRational ($($tt)*));
+        storage_types!(@type ($(#[$attr])*) @$M Complex32 ($($tt)*));
+        storage_types!(@type ($(#[$attr])*) @$M Complex64 ($($tt)*));
         storage_types!(@type ($(#[$attr])*) @$M f32 ($($tt)*));
         storage_types!(@type ($(#[$attr])*) @$M f64 ($($tt)*));
     };
@@ -175,6 +192,10 @@ macro_rules! storage_types {
         storage_types!(@type ($(#[$attr])*) @$M u64 ($($tt)*));
         storage_types!(@type ($(#[$attr])*) @$M u128 ($($tt)*));
         storage_types!(@type ($(#[$attr])*) @$M BigUint ($($tt)*));
+    };
+    (@type ($(#[$attr:meta])*) @$M:ident Complex ($($tt:tt)*)) => {
+        storage_types!(@type ($(#[$attr])*) @$M Complex32 ($($tt)*));
+        storage_types!(@type ($(#[$attr])*) @$M Complex64 ($($tt)*));
     };
     (@mod ($(#[$attr:meta])*) $M:ident, $V:ty; ($($tt:tt)*)) => {
         $(#[$attr])*
@@ -245,6 +266,8 @@ storage_type_types! {
     storage_type_rational32!("rational32", rational32, $crate::num::rational::Rational32);
     storage_type_rational64!("rational64", rational64, $crate::num::rational::Rational64);
     storage_type_bigrational!("bigrational", bigrational, $crate::num::BigRational);
+    storage_type_complex32!("complex32", complex32, $crate::num::complex::Complex32);
+    storage_type_complex64!("complex64", complex64, $crate::num::complex::Complex64);
     storage_type_f32!("f32", f32, f32);
     storage_type_f64!("f64", f64, f64);
 }
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -146,6 +146,56 @@ mod test_trait {
 
         impl super::super::Test for V {}
     }
+
+    storage_types! {
+        types: Complex;
+
+        use crate::num::Float;
+
+        // const EPSILON: VV = 64.0 * VV::epsilon(); //error[E0015]; calls in constants are limited...
+        const EPS_FACTOR: VV = 0.5;
+        const ULPS: u32 = 3;
+
+        impl super::super::Test for V {
+            /// Assert that `lhs` and `rhs` are exactly equal.
+            fn assert_eq(lhs: &Self, rhs: &Self) {
+                match (lhs.is_nan(), rhs.is_nan()) {
+                    (true, true) => {}
+                    _ => { assert_eq!(lhs, rhs); }
+                }
+            }
+
+            /// Assert that `lhs` and `rhs` are approximately equal for floating point types or
+            /// exactly equal for other types.
+            fn assert_approx_eq(lhs: &Self, rhs: &Self) {
+                match (lhs.is_nan(), rhs.is_nan()) {
+                    (true, true) => {}
+                    _ => {
+                        assert_ulps_eq!(lhs.re, rhs.re, epsilon = EPS_FACTOR * VV::epsilon(),
+                            max_ulps = ULPS);
+                        assert_ulps_eq!(lhs.im, rhs.im, epsilon = EPS_FACTOR * VV::epsilon(),
+                            max_ulps = ULPS);
+                    }
+                }
+            }
+
+            /// Exactly compare `lhs` and `rhs` and return the result.
+            fn eq(lhs: &Self, rhs: &Self) -> bool {
+                (lhs.is_nan() && rhs.is_nan())
+                    || lhs == rhs
+            }
+
+            /// Approximately compare `lhs` and `rhs` for floating point types or exactly compare
+            /// for other types and return the result.
+            fn approx_eq(lhs: &Self, rhs: &Self) -> bool {
+                (lhs.is_nan() && rhs.is_nan())
+                    || ulps_eq!(lhs.re, rhs.re,
+                        epsilon = EPS_FACTOR * VV::epsilon(), max_ulps = ULPS)
+                    || ulps_eq!(lhs.im, rhs.im,
+                        epsilon = EPS_FACTOR * VV::epsilon(), max_ulps = ULPS)
+            }
+        }
+    }
 }
 
 #[derive(Clone, Debug)]
@@ -190,6 +240,22 @@ mod a_struct {
                             break a;
                         }
                     },
+                }
+            }
+        }
+    }
+
+    storage_types! {
+        types: Complex;
+
+        use super::super::A;
+
+        impl quickcheck::Arbitrary for A<V> {
+            fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+                A {
+                    v: V::new(
+                        <VV as quickcheck::Arbitrary>::arbitrary(g),
+                        <VV as quickcheck::Arbitrary>::arbitrary(g)),
                 }
             }
         }
--- a/src/tests/quantity.rs
+++ b/src/tests/quantity.rs
@@ -173,71 +173,6 @@ storage_types! {
 
             a == x && b == y
         }
-
-        #[allow(trivial_casts)]
-        fn partial_cmp(l: A<V>, r: A<V>) -> bool {
-            let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
-            let a = (*l).partial_cmp(&(((*r).clone() / &km) * &km));
-            let b = ((*l).clone() / &km).partial_cmp(&((*r).clone() / &km));
-            let x = f::Length::new::<meter>((*l).clone()).partial_cmp(
-                &k::Length::new::<meter>((*r).clone()));
-            let y = k::Length::new::<meter>((*l).clone()).partial_cmp(
-                &f::Length::new::<meter>((*r).clone()));
-
-            a == x && b == y
-        }
-
-        #[allow(trivial_casts)]
-        fn lt(l: A<V>, r: A<V>) -> bool {
-            let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
-            let a = (*l).lt(&(((*r).clone() / &km) * &km));
-            let b = ((*l).clone() / &km).lt(&((*r).clone() / &km));
-            let x = f::Length::new::<meter>((*l).clone()).lt(
-                &k::Length::new::<meter>((*r).clone()));
-            let y = k::Length::new::<meter>((*l).clone()).lt(
-                &f::Length::new::<meter>((*r).clone()));
-
-            a == x && b == y
-        }
-
-        #[allow(trivial_casts)]
-        fn le(l: A<V>, r: A<V>) -> bool {
-            let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
-            let a = (*l).le(&(((*r).clone() / &km) * &km));
-            let b = ((*l).clone() / &km).le(&((*r).clone() / &km));
-            let x = f::Length::new::<meter>((*l).clone()).le(
-                &k::Length::new::<meter>((*r).clone()));
-            let y = k::Length::new::<meter>((*l).clone()).le(
-                &f::Length::new::<meter>((*r).clone()));
-
-            a == x && b == y
-        }
-
-        #[allow(trivial_casts)]
-        fn gt(l: A<V>, r: A<V>) -> bool {
-            let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
-            let a = (*l).gt(&(((*r).clone() / &km) * &km));
-            let b = ((*l).clone() / &km).gt(&((*r).clone() / &km));
-            let x = f::Length::new::<meter>((*l).clone()).gt(
-                &k::Length::new::<meter>((*r).clone()));
-            let y = k::Length::new::<meter>((*l).clone()).gt(
-                &f::Length::new::<meter>((*r).clone()));
-
-            a == x && b == y
-        }
-
-        #[allow(trivial_casts)]
-        fn ge(l: A<V>, r: A<V>) -> bool {
-            let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
-            let a = (*l).ge(&(((*r).clone() / &km) * &km));
-            let b = ((*l).clone() / &km).ge(&((*r).clone() / &km));
-            let x = f::Length::new::<meter>((*l).clone()).ge(
-                &k::Length::new::<meter>((*r).clone()));
-            let y = k::Length::new::<meter>((*l).clone()).ge(
-                &f::Length::new::<meter>((*r).clone()));
-
-            a == x && b == y
-        }
     }
 }
 
@@ -552,6 +487,86 @@ mod float {
                     &k::Length::new::<meter>(l).hypot(f::Length::new::<meter>(r)).get::<meter>());
 
                 fk && kf
+            }
+        }
+    }
+}
+
+#[cfg(feature = "autoconvert")]
+mod non_complex {
+    storage_types! {
+        // Everything BUT complex
+        types: PrimInt, Ratio, Float, BigInt, BigUint;
+
+        use crate::tests::*;
+
+        mod f { Q!(crate::tests, super::V); }
+        mod k { Q!(crate::tests, super::V, (kilometer, kilogram, kelvin)); }
+
+        quickcheck! {
+            #[allow(trivial_casts)]
+            fn partial_cmp(l: A<V>, r: A<V>) -> bool {
+                let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
+                let a = (*l).partial_cmp(&(((*r).clone() / &km) * &km));
+                let b = ((*l).clone() / &km).partial_cmp(&((*r).clone() / &km));
+                let x = f::Length::new::<meter>((*l).clone()).partial_cmp(
+                    &k::Length::new::<meter>((*r).clone()));
+                let y = k::Length::new::<meter>((*l).clone()).partial_cmp(
+                    &f::Length::new::<meter>((*r).clone()));
+
+                a == x && b == y
+            }
+
+            #[allow(trivial_casts)]
+            fn lt(l: A<V>, r: A<V>) -> bool {
+                let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
+                let a = (*l).lt(&(((*r).clone() / &km) * &km));
+                let b = ((*l).clone() / &km).lt(&((*r).clone() / &km));
+                let x = f::Length::new::<meter>((*l).clone()).lt(
+                    &k::Length::new::<meter>((*r).clone()));
+                let y = k::Length::new::<meter>((*l).clone()).lt(
+                    &f::Length::new::<meter>((*r).clone()));
+
+                a == x && b == y
+            }
+
+            #[allow(trivial_casts)]
+            fn le(l: A<V>, r: A<V>) -> bool {
+                let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
+                let a = (*l).le(&(((*r).clone() / &km) * &km));
+                let b = ((*l).clone() / &km).le(&((*r).clone() / &km));
+                let x = f::Length::new::<meter>((*l).clone()).le(
+                    &k::Length::new::<meter>((*r).clone()));
+                let y = k::Length::new::<meter>((*l).clone()).le(
+                    &f::Length::new::<meter>((*r).clone()));
+
+                a == x && b == y
+            }
+
+            #[allow(trivial_casts)]
+            fn gt(l: A<V>, r: A<V>) -> bool {
+                let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
+                let a = (*l).gt(&(((*r).clone() / &km) * &km));
+                let b = ((*l).clone() / &km).gt(&((*r).clone() / &km));
+                let x = f::Length::new::<meter>((*l).clone()).gt(
+                    &k::Length::new::<meter>((*r).clone()));
+                let y = k::Length::new::<meter>((*l).clone()).gt(
+                    &f::Length::new::<meter>((*r).clone()));
+
+                a == x && b == y
+            }
+
+            #[allow(trivial_casts)]
+            fn ge(l: A<V>, r: A<V>) -> bool {
+                let km: V = <kilometer as crate::Conversion<V>>::coefficient().value();
+                let a = (*l).ge(&(((*r).clone() / &km) * &km));
+                let b = ((*l).clone() / &km).ge(&((*r).clone() / &km));
+                let x = f::Length::new::<meter>((*l).clone()).ge(
+                    &k::Length::new::<meter>((*r).clone()));
+                let y = k::Length::new::<meter>((*l).clone()).ge(
+                    &f::Length::new::<meter>((*r).clone()));
+
+                a == x && b == y
             }
         }
     }
--- a/src/tests/system.rs
+++ b/src/tests/system.rs
@@ -200,37 +200,6 @@ storage_types! {
         }
 
         #[allow(trivial_casts)]
-        fn partial_cmp(l: A<V>, r: A<V>) -> bool {
-            (*l).partial_cmp(&*r)
-                == Length::new::<meter>((*l).clone()).partial_cmp(
-                    &Length::new::<meter>((*r).clone()))
-        }
-
-        #[allow(trivial_casts)]
-        fn lt(l: A<V>, r: A<V>) -> bool {
-            (*l).lt(&*r)
-                == Length::new::<meter>((*l).clone()).lt(&Length::new::<meter>((*r).clone()))
-        }
-
-        #[allow(trivial_casts)]
-        fn le(l: A<V>, r: A<V>) -> bool {
-            (*l).le(&*r)
-                == Length::new::<meter>((*l).clone()).le(&Length::new::<meter>((*r).clone()))
-        }
-
-        #[allow(trivial_casts)]
-        fn gt(l: A<V>, r: A<V>) -> bool {
-            (*l).gt(&*r)
-                == Length::new::<meter>((*l).clone()).gt(&Length::new::<meter>((*r).clone()))
-        }
-
-        #[allow(trivial_casts)]
-        fn ge(l: A<V>, r: A<V>) -> bool {
-            (*l).ge(&*r)
-                == Length::new::<meter>((*l).clone()).ge(&Length::new::<meter>((*r).clone()))
-        }
-
-        #[allow(trivial_casts)]
         fn rem(l: A<V>, r: A<V>) -> TestResult {
             if *r == V::zero() {
                 return TestResult::discard();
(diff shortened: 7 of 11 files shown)


Cargo.toml
```diff
@@ -36,6 +36,7 @@ members = [
 num-traits = { version = "0.2", default-features = false }
 num-rational = { version = "0.4", optional = true, default-features = false }
 num-bigint = { version = "0.4", optional = true, default-features = false, features = ["std"] }
+num-complex = { version = "0.4", optional = true, default-features = false }
```
iliekturtles, line 39: I think `std` is required for `Float` types based on the skim of `num-complex` documentation that I did.

```suggestion
num-complex = { version = "0.4", optional = true, default-features = false, features = ["std"] }
```

iliekturtles (commented):
Thanks for the PR! I did a very brief review and you're on the right track. I'll review in more detail tomorrow and see if I can assist with some of the trait implementations that are still missing.

TobTobXX:
Oh, is it a problem if the types don't implement `PartialOrd`? As far as I can tell, `num::complex::Complex<T>` doesn't implement it (which does make some kind of mathematical sense). And as far as I can tell, `num::Num` doesn't automatically implement `PartialOrd`.

It seems to be a problem because `uom::ConversionFactor<V>` can only be implemented for `PartialOrd` types: https://github.com/iliekturtles/uom/blob/fe7f0b38ca1daffe67dd60b3adb1676766f0caec/src/lib.rs#L441

Kinda crazy how some intuitive promises are broken by math.

src/storage_types.rs
```diff
@@ -125,6 +132,8 @@ macro_rules! storage_types {
         storage_types!(@type ($(#[$attr])*) @$M Rational32 ($($tt)*));
         storage_types!(@type ($(#[$attr])*) @$M Rational64 ($($tt)*));
         storage_types!(@type ($(#[$attr])*) @$M BigRational ($($tt)*));
+        storage_types!(@type ($(#[$attr])*) @$M Complex32 ($($tt)*));
+        storage_types!(@type ($(#[$attr])*) @$M Complex64 ($($tt)*));
```
iliekturtles, line 136: This change adds a new type alias, `VV` for complex underlying storage types. e.g. `f32` for `Complex32` and `f64` for `Complex64`.

```suggestion
        storage_types!(@type ($(#[$attr])*) @$M Complex32 (#[allow(dead_code)]
            pub type VV = f32;
            $($tt)*));
        storage_types!(@type ($(#[$attr])*) @$M Complex64 (#[allow(dead_code)]
            pub type VV = f64;
            $($tt)*));
```

src/lib.rs
```diff
@@ -636,6 +649,31 @@ storage_types! {
...
+        fn powi(self, e: i32) -> Self {
+            self.powi(e)
+        }
+
+        #[inline(always)]
+        fn value(self) -> V {
+            self
+        }
+    }
+}
```
iliekturtles, line 675: Use `Complex<T>`'s underlying type when implementing `Conversion<V>` and `ConversionFactor<V>`. These changes essentially say that the conversion factor type used to convert between units for `Complex<T>` is `T`. See the other suggestion where the definition for `VV` is added.

EDIT: The diff doesn't come out very good. The main changes are `type T = VV;` and `impl crate::ConversionFactor<V> for VV {`

```suggestion
storage_types! {
    types: Complex;

    impl crate::Conversion<V> for V {
        type T = VV;

        #[inline(always)]
        fn constant(op: crate::ConstantOp) -> Self::T {
            match op {
                crate::ConstantOp::Add => -<Self::T as crate::num::Zero>::zero(),
                crate::ConstantOp::Sub => <Self::T as crate::num::Zero>::zero(),
            }
        }

        #[inline(always)]
        fn conversion(&self) -> Self::T {
            *self
        }
    }

    impl crate::ConversionFactor<V> for VV {
        #[inline(always)]
        fn powi(self, e: i32) -> Self {
            self.powi(e)
        }

        #[inline(always)]
        fn value(self) -> V {
            self
        }
    }
}
```

iliekturtles (commented):
I believe the `PartialOrd` issue can be resolved by using `Complex<T>`'s underlying type as the `ConversionFactor<V>` type. See the two suggestions in this review for the changes that I believe are needed. I haven't attempted to compile so additional changes are likely needed.

TobTobXX:
It may have worked (I'm still getting errors, but I think unrelated ones, I'll work on them after). Just to check my thought process, could you confirm/correct this (applies to `f64` too):

1. uom needs to know how to convert `Complex32` numbers (eg. between _12i+3 km_ to _m_)
2. Desired behavior is just scaling (eg. _12i+3 km_ == _1e3 * (12i+3) m_)
3. Because having this scaling factor be a complex number has some complications (ie. [prev. errors](https://github.com/iliekturtles/uom/pull/287#issuecomment-1064591768)), we want the conversion value to be a `f32` (conversion factors don't need an imaginary part anyway).
4. I need to implement `Conversion<Complex32>` to define how conversions should work.
5. The implementation `ConversionFactor<Complex32> for f32` declares that `Complex32` numbers can be converted with `f32` numbers.
6. `Conversion<Complex32, T = f32>::conversion()` should return the norm of the complex number, since it can be multiplied by this factor again to get the original number.
7. `ConversionFactor<Complex32>::value()` for `Self = f32` should return _(self + 0i)_. Because _12i+3km_ == _(1e3+0i) * (12i+3) m_ (so it is basically just glue to satisfy the type system).

Also, I have no idea what that `Conversion::constant()` function is doing with this evil `0.0 != -0.0` stuff. I'll just copy your suggestion... (Is it even needed if I overwrite the `Conversion::conversion()` fn? Or should I overwrite `Conversion::coefficient() instead?)

Thank you for your patience and help :hugs:

TobTobXX:
It seems to work now btw :D

iliekturtles:
Excellent! I started the build and will review in more detail later today.

src/storage_types.rs
```diff
@@ -7,15 +7,16 @@
 /// * `$T`: Types to generate a module for. Accepts all underlying storage types along with a number
 ///   of different categories:
 ///   * `All`: `usize`, `u8`, `u16`, `u32`, `u64`, `u128`, `isize`, `i8`, `i16`, `i32`, `i64`,
-///     `i128`, `BigInt`, `BigUint`, `Rational`, `Rational32`, `Rational64`, `BigRational`, `f32`,
-///     and `f64`.
+///     `i128`, `BigInt`, `BigUint`, `Rational`, `Rational32`, `Rational64`, `BigRational`,
```
iliekturtles, line 10: Remove trailing space.

```suggestion
///     `i128`, `BigInt`, `BigUint`, `Rational`, `Rational32`, `Rational64`, `BigRational`,
```

iliekturtles (commented):
`tests::Test` needs to be implemented. See the relevant code for `f32`/`f64` below. You'll want to use similar approx comparisons for the real part. I assume you'll want approx comparisons for the imaginary part.

https://github.com/iliekturtles/uom/blob/fe7f0b38ca1daffe67dd60b3adb1676766f0caec/src/tests/mod.rs#L99-L142

TobTobXX:
I started some work. A lot of code duplication and still quite a few errors, but could you look over these points:

src/tests/system.rs
```diff
@@ -607,3 +607,148 @@ mod fixed {
...
+            ).is_normal());
+            assert!(!Length::new::<meter>(
+                V::new(VV::zero(), VV::zero())
+            ).is_normal());
+            assert!(!Length::new::<meter>(
+                V::new(VV::nan(), VV::nan())
+            ).is_normal());
+            assert!(!Length::new::<meter>(
+                V::new(VV::infinity(), VV::infinity())
+            ).is_normal());
```
TobTobXX, line 647: This only partially makes sense for complex numbers, as you'd have to test eg. whether the complex number is_infinite() when _either_ the real or imaginary part is infinite. Should I leave it like this, expand to cover all cases or just remove it?

src/tests/system.rs
```diff
@@ -607,3 +607,148 @@ mod fixed {
...
+            fn max(l: A<V>, r: A<V>) -> bool {
+                Test::eq(&Length::new::<meter>(l.max(*r)),
+                    &Length::new::<meter>(*l).max(Length::new::<meter>(*r)))
+            }
+
+            #[allow(trivial_casts)]
+            fn min(l: A<V>, r: A<V>) -> bool {
+                Test::eq(&Length::new::<meter>(l.min(*r)),
+                    &Length::new::<meter>(*l).min(Length::new::<meter>(*r)))
+            }
```
TobTobXX, line 751: Exactly duplicated from the `quickcheck!` block above for `Float`. Can this duplication be reduced?

iliekturtles: You can extract the duplicated code into a `storage_types!` block with `types: Complex, Float;`. For the `is_*` functions that are currently slightly different you can just use `Complex`'s `is_*` implementation. e.g. [`Complex::is_normal`](https://docs.rs/num-complex/0.4.0/src/num_complex/lib.rs.html#590) is already doing `v.re.is_normal() && v.im.is_normal()`

One thing to keep in mind when doing this is that `storage_types!` always generates modules with a name of the type. This means you can't execute `storage_types!` for the same types in the same parent module and may need to introduce a new child module.

iliekturtles:
Approved the build again so that you'll get test results. I'm hoping to have time tomorrow to do an in depth review.

iliekturtles:
Looks like the issue is that `Complex<T>` implements the methods when `T: FloatCore`, but `Complex<T>` itself does not implement `FloatCore` (or `Float`). This may require a `storage_types!` call in `system.rs` or `quantity.rs`. I'll check in more detail during my review.

iliekturtles:
Despite my best intentions I just didn't get to doing a deeper review. Hopefully this weekend or early next week.

TobTobXX:
no problem, I'm not in a hurry.

src/lib.rs
```diff
@@ -636,6 +649,32 @@ storage_types! {
...
+            self.powi(e)
+        }
+        #[inline(always)]
+        fn value(self) -> V {
+            // Conversion by scaling (multiplication with only real number)
+            // Scaling a normalized number yields the original number again.
+            V::new(self, 0.0)
+        }
+    }
+}
```
iliekturtles, line 676: * Some minor formatting.
 *  `fn constant` is needed to ensure zero-cost code generation. See documentation on the trait for details.

```suggestion
storage_types! {
    types: Complex;

    impl crate::Conversion<V> for V {
        type T = VV;

        #[inline(always)]
        fn constant(op: crate::ConstantOp) -> Self::T {
            match op {
                crate::ConstantOp::Add => -<Self::T as crate::num::Zero>::zero(),
                crate::ConstantOp::Sub => <Self::T as crate::num::Zero>::zero(),
            }
        }

        #[inline(always)]
        fn conversion(&self) -> Self::T {
            // Conversion factor is the norm of the number. Scaling with length again yields the
            // same number.
            self.norm()
        }
    }

    impl crate::ConversionFactor<V> for VV {
        #[inline(always)]
        fn powi(self, e: i32) -> Self {
            self.powi(e)
        }

        #[inline(always)]
        fn value(self) -> V {
            // Conversion by scaling (multiplication with only real number). Scaling a normalized
            // number yields the original number again.
            V::new(self, 0.0)
        }
    }
}
```

iliekturtles (commented):
Here is the review for the first four files. I haven't gotten through the remaining files and the `the trait bound `Complex<f32>: num_traits::Float` is not satisfied` error still exists. Will try to get through the rest tomorrow.

.github/workflows/ci-full-test-suite.yml
```diff
@@ -60,4 +60,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --verbose --no-run --no-default-features --features "autoconvert usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 bigint biguint rational rational32 rational64 bigrational f32 f64 std use_serde"
+          args: --verbose --no-run --no-default-features --features "autoconvert usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 bigint biguint rational rational32 rational64 bigrational complex32 complex64 f32 f64 std use_serde"
```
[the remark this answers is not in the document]
TobTobXX: I added `complex32` AND `complex64`. Correct me if wrong.

TobTobXX:
Will address the last review later. Rebased, will squash sometime later too...

src/lib.rs
```diff
@@ -214,6 +216,10 @@ pub extern crate num_bigint;
 #[cfg(any(feature = "rational-support", feature = "bigint-support"))]
 pub extern crate num_rational;
 
+#[doc(hidden)]
+#[cfg(feature = "compled-support")]
```
iliekturtles, line 220:
```suggestion
#[cfg(feature = "complex-support")]
```

src/tests/mod.rs
```diff
@@ -146,6 +146,58 @@ mod test_trait {
...
+                match (lhs.is_nan(), rhs.is_nan()) {
+                    (true, true) => {}
+                    _ => { assert_eq!(lhs, rhs); }
+                }
+            }
+
+            /// Assert that `lhs` and `rhs` are approximately equal for floating point types or
+            /// exactly equal for other types.
+            fn assert_approx_eq(lhs: &Self, rhs: &Self) {
+                use approx::assert_ulps_eq;
```
iliekturtles, line 171: No longer needed once `extern crate approx;` fix in `lib.rs` is applied.

```suggestion
```

src/tests/mod.rs
```diff
@@ -146,6 +146,58 @@ mod test_trait {
 
         impl super::super::Test for V {}
     }
+
+    storage_types! {
+        types: Complex;
+
+        use crate::num::Float;
+
+        // const EPSILON: V = 64.0 * V::epsilon(); //error[E0015]; calls in constants are limited...
```
iliekturtles, line 155:
```suggestion
        // const EPSILON: VV = 64.0 * VV::epsilon(); //error[E0015]; calls in constants are limited...
```

src/storage_types.rs
```diff
@@ -100,6 +101,12 @@ macro_rules! storage_types {
     (@type ($(#[$attr:meta])*) @$M:ident BigRational ($($tt:tt)*)) => {
         storage_type_bigrational!(($(#[$attr])*) @$M ($($tt)*));
     };
+    (@type ($(#[$attr:meta])*) @$M:ident Complex32 ($($tt:tt)*)) => {
+        storage_type_complex32!(($(#[$attr])*) @$M ($($tt)*));
+    };
+    (@type ($(#[$attr:meta])*) @$M:ident Complex64 ($($tt:tt)*)) => {
+        storage_type_complex64!(($(#[$attr])*) @$M ($($tt)*));
+    };
```
iliekturtles, line 109: Move `VV` type aliases to this location.

```suggestion
    (@type ($(#[$attr:meta])*) @$M:ident Complex32 ($($tt:tt)*)) => {
        storage_type_complex32!(($(#[$attr])*) @$M (
            /// Inner storage type.
            #[allow(dead_code)]
            pub type VV = f32;
            $($tt)*));
    };
    (@type ($(#[$attr:meta])*) @$M:ident Complex64 ($($tt:tt)*)) => {
        storage_type_complex64!(($(#[$attr])*) @$M (
            /// Inner storage type.
            #[allow(dead_code)]
            pub type VV = f64;
            $($tt)*));
    };
```

src/tests/mod.rs
```diff
@@ -146,6 +146,58 @@ mod test_trait {
...
+            /// Exactly compare `lhs` and `rhs` and return the result.
+            fn eq(lhs: &Self, rhs: &Self) -> bool {
+                (lhs.is_nan() && rhs.is_nan())
+                    || lhs == rhs
+            }
+
+            /// Approximately compare `lhs` and `rhs` for floating point types or exactly compare
+            /// for other types and return the result.
+            fn approx_eq(lhs: &Self, rhs: &Self) -> bool {
+                use approx::ulps_eq;
```
iliekturtles, line 192:
```suggestion
```

src/tests/mod.rs
```diff
@@ -194,6 +246,23 @@ mod a_struct {
...
+        use super::super::A;
+
+        impl quickcheck::Arbitrary for A<V> {
+            fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+                A {
+                    v: V::new(
+                        <VV as quickcheck::Arbitrary>::arbitrary(g),
+                        <VV as quickcheck::Arbitrary>::arbitrary(g),
+                    ),
+                }
```
iliekturtles, line 262:
```suggestion
                A {
                    v: V::new(
                        <VV as quickcheck::Arbitrary>::arbitrary(g),
                        <VV as quickcheck::Arbitrary>::arbitrary(g)),
                }
```

iliekturtles (commented):
With the changes in this review and if you temporarily comment out all of the methods like `is_normal` that rely on `Float`/`FloatCore` the crate will compile and run tests! Some of the tests are failing and I haven't figured out what the solution for the `Float`/`FloatCore` is going to be yet.

I'm also noticing that a lot of the `Complex` `storage_types!` blocks are a copy-and-paste of the `Float` blocks with `V` replaced with `VV`. I don't want to block acceptance of this PR on reducing this duplicate code. Please post If you have any thoughts about the best way to reduce this duplication.

Add new features to `extern crate approx;` statement:

```diff
- #[cfg(all(test, any(feature = "f32", feature = "f64")))]
+ #[cfg(all(test, any(feature = "f32", feature = "f64", feature = "complex32", feature = "complex64")))]
  #[macro_use]
  extern crate approx;
```

Add to `src/tests/asserts.rs`:

```rust
storage_types! {
    types: Complex;

    use super::*;

    assert_impl_all!(Quantity<Q<Z0, Z0, Z0>, U<V>, V>:
        Clone, Copy, Debug, PartialEq, Send, Sync, Unpin);
    #[cfg(feature = "std")]
    assert_impl_all!(Quantity<Q<Z0, Z0, Z0>, U<V>, V>:
        RefUnwindSafe, UnwindSafe);
    assert_not_impl_any!(Quantity<Q<Z0, Z0, Z0>, U<V>, V>:
        Binary, Display, Eq, Hash, LowerExp, LowerHex, Octal, Ord, PartialOrd, UpperExp, UpperHex);
    assert_impl_all!(QuantityArguments<Q<Z0, Z0, Z0>, U<V>, V, meter>:
        Clone, Copy, Debug, Display, LowerExp, Send, Sync, Unpin, UpperExp);
    #[cfg(feature = "std")]
    assert_impl_all!(QuantityArguments<Q<Z0, Z0, Z0>, U<V>, V, meter>:
        RefUnwindSafe, UnwindSafe);
    assert_not_impl_any!(QuantityArguments<Q<Z0, Z0, Z0>, U<V>, V, meter>:
        Binary, Eq, Hash, LowerHex, Octal, Ord, PartialEq, PartialOrd, UpperHex);
}
```

.github/workflows/ci-full-test-suite.yml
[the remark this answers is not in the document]
iliekturtles: Line 57 just needs one. I'd go with `Complex32`. Line 63 should have both.

TobTobXX:
Ok, sorry I got a bit delayed on this, but all should be resolved now.

I commented out the tests where the weird issues with `T: Float` and `T: FloatCore` appear, and added a notice.

And about the duplication with `VV` instead of `V`: Nope, got no idea. I'm no macro god like you. :joy:

iliekturtles:
Excellent! I'll try to get to reviewing the latest changes this week.

src/tests/quantity.rs
```diff
@@ -556,3 +491,82 @@ mod float {
...
+
+mod non_complex {
+    storage_types! {
+        // Everything BUT complex
+        types: PrimInt, Ratio, Float, BigInt, BigUint;
+
+        use crate::tests::*;
+
+        mod f { Q!(crate::tests, super::V); }
+        mod k { Q!(crate::tests, super::V, (kilometer, kilogram, kelvin)); }
```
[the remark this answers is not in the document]
iliekturtles: The tests are for using a system of units with a different set of base units (km, kg, K) instead of (m, kg, K). There may be a fix for this in #229. I think that PR is pretty close and I just need to push it over the line.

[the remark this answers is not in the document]
iliekturtles: #229 was unrelated. A feature gate was missing!

```diff
diff --git a/src/tests/quantity.rs b/src/tests/quantity.rs
index a0f1383..90bd9b7 100644
--- a/src/tests/quantity.rs
+++ b/src/tests/quantity.rs
@@ -502,6 +502,7 @@ mod non_complex {
         mod f { Q!(crate::tests, super::V); }
         mod k { Q!(crate::tests, super::V, (kilometer, kilogram, kelvin)); }
 
+        #[cfg(feature = "autoconvert")]
         quickcheck! {
             #[allow(trivial_casts)]
             fn partial_cmp(l: A<V>, r: A<V>) -> bool {
```

src/lib.rs
```diff
@@ -645,6 +659,38 @@ storage_types! {
     }
 }
 
+storage_types! {
+    types: Complex;
+    impl crate::Conversion<V> for V {
+        type T = VV;
```
iliekturtles, line 665:
```suggestion
        type T = VV;
        
```

src/lib.rs
```diff
@@ -645,6 +659,38 @@ storage_types! {
...
+        #[inline(always)]
+        fn constant(op: crate::ConstantOp) -> Self::T {
+            match op {
+                crate::ConstantOp::Add => -<Self::T as crate::num::Zero>::zero(),
+                crate::ConstantOp::Sub => <Self::T as crate::num::Zero>::zero(),
+            }
+        }
+        #[inline(always)]
```
iliekturtles, line 673:
```suggestion
        }
        
        #[inline(always)]
```

src/lib.rs
```diff
@@ -645,6 +659,38 @@ storage_types! {
...
+        fn conversion(&self) -> Self::T {
+            // Conversion factor is the norm of the number. Scaling with length again yields the
+            // same number.
+            self.norm()
+        }
+    }
+    impl crate::ConversionFactor<V> for VV {
```
iliekturtles, line 680:
```suggestion
    }
    
    impl crate::ConversionFactor<V> for VV {
```

src/lib.rs
```diff
@@ -645,6 +659,38 @@ storage_types! {
...
+        #[inline(always)]
+        fn powi(self, e: i32) -> Self {
+            self.powi(e)
+        }
+        #[inline(always)]
```
iliekturtles, line 685:
```suggestion
        }
        
        #[inline(always)]
```

src/tests/system.rs
```diff
@@ -607,3 +576,195 @@ mod fixed {
...
+
+// Somehow Complex<f32|f64> don't implement these methods, even though they
+// should, bc f32 / f64 implement Float and FloatCore:
+//  - https://docs.rs/num/latest/num/struct.Complex.html#impl-5
+//  - https://docs.rs/num/latest/num/struct.Complex.html#impl-4
+//
+// See discussion here:
+// https://github.com/iliekturtles/uom/pull/287#pullrequestreview-922005933
+/*
+mod complex {
```
iliekturtles, line 588: I spent some time looking at this tonight and I believe what we'll need to do is explicitly implement these functions for `f32`, `f64`, `Complex32`, and `Complex64` using the `storage_types!` macro. `Complex<T>` doesn't implement `Float`/`FloatCore`, but it does implement these methods when the underlying type, `T`, implements `FloatCore`.

```rust
storage_types! {
    types: Float, Complex;

    impl<...> Quantity<..., V /*V type alias defined by storage_types!*/> {
        fn is_infinite(self) -> bool { self.value.is_infinite() }
    }
}

// I don't think a separate storage_types! is necessary. The code should be able to be the same.
//storage_types! {
//    types: Complex;
//    
//    impl<...> Quantity<..., V> {
//        fn is_infinite(...) -> ... { ... }
//    }
}

File: src/tests/quantity.rs (hunk 15 above)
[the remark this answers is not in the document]
Comment (iliekturtles) [reply src/tests/quantity.rs:503]: `#[cfg(...)]` can actually be moved up on top of the mod statement.

File: src/tests/system.rs (hunk 20 above)
Comment (TobTobXX) [reply src/tests/system.rs:588]: I don't completely understand why I should implement these methods on `f32` and `f64` too, num implements them for these two just fine, doesn't it?

As for doing it for the complexes... it's weird. Before the compiler was complaining that it didn't find an implementation for `is_infinite` et al., but now when I implement it with the following patch, it tells me it has duplicate implementation:

```diff
diff --git a/src/tests/system.rs b/src/tests/system.rs
index 1355f80..cad6de9 100644
--- a/src/tests/system.rs
+++ b/src/tests/system.rs
@@ -584,7 +584,6 @@ mod fixed {
 //
 // See discussion here:
 // https://github.com/iliekturtles/uom/pull/287#pullrequestreview-922005933
-/*
 mod complex {
     storage_types! {
         types: Complex;
@@ -593,6 +592,11 @@ mod complex {
 
         Q!(crate::tests, V);
 
+        impl<D, U> Quantity<D, U, V /*V type alias defined by storage_types!*/> {
+            fn is_infinite(self) -> bool { self.value.is_infinite() }
+            fn is_finite(self) -> bool { self.value.is_finite() }
+        }
+
         #[test]
         fn fp_categories() {
             let float_categories = vec![
@@ -724,7 +728,6 @@ mod complex {
         }
     }
 }
-*/
 
 mod non_complex {
     storage_types! {

```

```
error[E0592]: duplicate definitions with name `is_infinite`
    |
   ::: src/tests/system.rs:596:13
    |
596 |               fn is_infinite(self) -> bool { self.value.is_infinite() }
    |               ---------------------------- other definition for `is_infinite`
    |
   ::: src/tests/mod.rs:55:1
    |
55  | / system! {
56  | |     quantities: Q {
57  | |         length: meter, L;
58  | |         mass: kilogram, M;
...   |
65  | |     }
66  | | }
   --> src/system.rs:613:13
    |
613 | /             pub fn is_infinite(self) -> bool
614 | |             where
615 | |                 V: $crate::num::Float,
    | |______________________________________^ duplicate definitions for `is_infinite`
    |
    = note: this error originates in the macro `system` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Maybe I am also implementing it at the wrong location?

iliekturtles: I think I see where the duplicates are and will take a look in the next few days!

iliekturtles:
We're getting closer. I got all the tests compiling by moving the general function implementations `where V: Float` into explicit implementations for each concrete type `V`. There are now some test failures that I didn't get a chance to review. Could you review, and possibly apply, the patch below?

```diff
diff --git a/src/system.rs b/src/system.rs
index 6324553..016f714 100644
--- a/src/system.rs
+++ b/src/system.rs
@@ -596,47 +596,6 @@ macro_rules! system {
             U: Units<V> + ?Sized,
             V: $crate::num::Num + $crate::Conversion<V>,
         {
-            /// Returns `true` if this value is `NAN` and `false` otherwise.
-            #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
-            #[inline(always)]
-            pub fn is_nan(self) -> bool
-            where
-                V: $crate::num::Float,
-            {
-                self.value.is_nan()
-            }
-
-            /// Returns `true` if this value is positive infinity or negative infinity and
-            /// `false` otherwise.
-            #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
-            #[inline(always)]
-            pub fn is_infinite(self) -> bool
-            where
-                V: $crate::num::Float,
-            {
-                self.value.is_infinite()
-            }
-
-            /// Returns `true` if this number is neither infinite nor `NAN`.
-            #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
-            #[inline(always)]
-            pub fn is_finite(self) -> bool
-            where
-                V: $crate::num::Float,
-            {
-                self.value.is_finite()
-            }
-
-            /// Returns `true` if the number is neither zero, infinite, subnormal, or `NAN`.
-            #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
-            #[inline(always)]
-            pub fn is_normal(self) -> bool
-            where
-                V: $crate::num::Float,
-            {
-                self.value.is_normal()
-            }
-
             /// Returns the floating point category of the number. If only one property is
             /// going to be tested, it is generally faster to use the specific predicate
             /// instead.
@@ -649,43 +608,6 @@ macro_rules! system {
             }
 
             std! {
-            /// Takes the cubic root of a number.
-            ///
-            #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
-            #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
-            /// # use uom::si::f32::*;
-            /// # use uom::si::volume::cubic_meter;
-            /// let l: Length = Volume::new::<cubic_meter>(8.0).cbrt();
-            /// ```
-            ///
-            /// The input type must have dimensions divisible by three:
-            ///
-            #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust,compile_fail")]
-            #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
-            /// # use uom::si::f32::*;
-            /// # use uom::si::area::square_meter;
-            /// // error[E0271]: type mismatch resolving ...
-            /// let r = Area::new::<square_meter>(8.0).cbrt();
-            /// ```
-            #[inline(always)]
-            pub fn cbrt(
-                self
-            ) -> Quantity<
-                $quantities<$($crate::typenum::PartialQuot<D::$symbol, $crate::typenum::P3>),+>,
-                U, V>
-            where
-                $(D::$symbol: $crate::lib::ops::PartialDiv<$crate::typenum::P3>,
-                <D::$symbol as $crate::lib::ops::PartialDiv<$crate::typenum::P3>>::Output: $crate::typenum::Integer,)+
-                D::Kind: $crate::marker::Div,
-                V: $crate::num::Float,
-            {
-                Quantity {
-                    dimension: $crate::lib::marker::PhantomData,
-                    units: $crate::lib::marker::PhantomData,
-                    value: self.value.cbrt(),
-                }
-            }
-
             autoconvert! {
             /// Calculates the length of the hypotenuse of a right-angle triangle given the legs.
             #[inline(always)]
@@ -769,39 +691,6 @@ macro_rules! system {
                 self.value.is_sign_negative()
             }
 
-            std! {
-            /// Fused multiply-add. Computes `(self * a) + b` with only one rounding error.
-            /// This produces a more accurate result with better performance than a separate
-            /// multiplication operation followed by an add.
-            ///
-            /// ## Generic Parameters
-            /// * `Da`: Dimension for parameter `a`.
-            /// * `Ua`: Base units for parameter `a`.
-            /// * `Ub`: Base units for parameter `b`.
-            #[inline(always)]
-            pub fn mul_add<Da, Ua, Ub>(
-                self,
-                a: Quantity<Da, Ua, V>,
-                b: Quantity<$quantities<$($crate::typenum::Sum<D::$symbol, Da::$symbol>),+>, Ub, V>,
-            ) -> Quantity<$quantities<$($crate::typenum::Sum<D::$symbol, Da::$symbol>),+>, U, V>
-            where
-                $(D::$symbol: $crate::lib::ops::Add<Da::$symbol>,
-                <D::$symbol as $crate::lib::ops::Add<Da::$symbol>>::Output: $crate::typenum::Integer,)+
-                D::Kind: $crate::marker::Mul,
-                V: $crate::num::Float,
-                Da: Dimension + ?Sized,
-                Da::Kind: $crate::marker::Mul,
-                Ua: Units<V> + ?Sized,
-                Ub: Units<V> + ?Sized,
-            {
-                // (self * a) + b
-                Quantity {
-                    dimension: $crate::lib::marker::PhantomData,
-                    units: $crate::lib::marker::PhantomData,
-                    value: self.value.mul_add(a.value, b.value),
-                }
-            }}
-
             /// Takes the reciprocal (inverse) of a number, `1/x`.
             ///
             #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
@@ -827,76 +716,6 @@ macro_rules! system {
                 }
             }
 
-            std! {
-            /// Raises a quantity to an integer power.
-            ///
-            #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
-            #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
-            /// # use uom::si::f32::*;
-            /// # use uom::si::length::meter;
-            /// use uom::typenum::P2;
-            ///
-            /// let a: Area = Length::new::<meter>(3.0).powi(P2::new());
-            /// ```
-            ///
-            /// ## Generic Parameters
-            /// * `E`: `typenum::Integer` power.
-            #[inline(always)]
-            pub fn powi<E>(
-                self, _e: E
-            ) -> Quantity<$quantities<$($crate::typenum::Prod<D::$symbol, E>),+>, U, V>
-            where
-                $(D::$symbol: $crate::lib::ops::Mul<E>,
-                <D::$symbol as $crate::lib::ops::Mul<E>>::Output: $crate::typenum::Integer,)+
-                D::Kind: $crate::marker::Mul,
-                E: $crate::typenum::Integer,
-                V: $crate::num::Float,
-            {
-                Quantity {
-                    dimension: $crate::lib::marker::PhantomData,
-                    units: $crate::lib::marker::PhantomData,
-                    value: self.value.powi(E::to_i32()),
-                }
-            }
-
-            /// Takes the square root of a number. Returns `NAN` if `self` is a negative
-            /// number.
-            ///
-            #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
-            #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
-            /// # use uom::si::f32::*;
-            /// # use uom::si::area::square_meter;
-            /// let l: Length = Area::new::<square_meter>(4.0).sqrt();
-            /// ```
-            ///
-            /// The input type must have dimensions divisible by two:
-            ///
-            #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust,compile_fail")]
-            #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
-            /// # use uom::si::f32::*;
-            /// # use uom::si::length::meter;
-            /// // error[E0271]: type mismatch resolving ...
-            /// let r = Length::new::<meter>(4.0).sqrt();
-            /// ```
-            #[inline(always)]
-            pub fn sqrt(
-                self
-            ) -> Quantity<
-                $quantities<$($crate::typenum::PartialQuot<D::$symbol, $crate::typenum::P2>),+>,
-                U, V>
-            where
-                $(D::$symbol: $crate::typenum::PartialDiv<$crate::typenum::P2>,
-                <D::$symbol as $crate::typenum::PartialDiv<$crate::typenum::P2>>::Output: $crate::typenum::Integer,)+
-                D::Kind: $crate::marker::Div,
-                V: $crate::num::Float,
-            {
-                Quantity {
-                    dimension: $crate::lib::marker::PhantomData,
-                    units: $crate::lib::marker::PhantomData,
-                    value: self.value.sqrt(),
-                }
-            }}
-
             /// Returns the maximum of the two quantities.
             #[inline(always)]
             pub fn max(self, other: Self) -> Self
@@ -924,6 +743,194 @@ macro_rules! system {
             }
         }
 
+        // Explicitly definte floating point methods for float and complex storage types.
+        // `Complex<T>` doesn't implement `Float`/`FloatCore`, but it does implement these methods
+        // when the underlying type, `T`, implements `FloatCore`.
+        mod float {
+            storage_types! {
+                types: Float, Complex;
+
+                use super::super::*;
+
+                impl<D, U> Quantity<D, U, V>
+                where
+                    D: Dimension + ?Sized,
+                    U: Units<V> + ?Sized,
+                {
+                    /// Returns `true` if this value is `NAN` and `false` otherwise.
+                    #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
+                    #[inline(always)]
+                    pub fn is_nan(self) -> bool
+                    {
+                        self.value.is_nan()
+                    }
+
+                    /// Returns `true` if this value is positive infinity or negative infinity and
+                    /// `false` otherwise.
+                    #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
+                    #[inline(always)]
+                    pub fn is_infinite(self) -> bool
+                    {
+                        self.value.is_infinite()
+                    }
+
+                    /// Returns `true` if this number is neither infinite nor `NAN`.
+                    #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
+                    #[inline(always)]
+                    pub fn is_finite(self) -> bool
+                    {
+                        self.value.is_finite()
+                    }
+
+                    /// Returns `true` if the number is neither zero, infinite, subnormal, or `NAN`.
+                    #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
+                    #[inline(always)]
+                    pub fn is_normal(self) -> bool
+                    {
+                        self.value.is_normal()
+                    }
+
+                    std! {
+                    /// Takes the cubic root of a number.
+                    ///
+                    #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
+                    #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
+                    /// # use uom::si::f32::*;
+                    /// # use uom::si::volume::cubic_meter;
+                    /// let l: Length = Volume::new::<cubic_meter>(8.0).cbrt();
+                    /// ```
+                    ///
+                    /// The input type must have dimensions divisible by three:
+                    ///
+                    #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust,compile_fail")]
+                    #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
+                    /// # use uom::si::f32::*;
+                    /// # use uom::si::area::square_meter;
+                    /// // error[E0271]: type mismatch resolving ...
+                    /// let r = Area::new::<square_meter>(8.0).cbrt();
+                    /// ```
+                    #[inline(always)]
+                    pub fn cbrt(
+                        self
+                    ) -> Quantity<
+                        $quantities<$($crate::typenum::PartialQuot<D::$symbol, $crate::typenum::P3>),+>,
+                        U, V>
+                    where
+                        $(D::$symbol: $crate::lib::ops::PartialDiv<$crate::typenum::P3>,
+                        <D::$symbol as $crate::lib::ops::PartialDiv<$crate::typenum::P3>>::Output: $crate::typenum::Integer,)+
+                        D::Kind: $crate::marker::Div,
+                    {
+                        Quantity {
+                            dimension: $crate::lib::marker::PhantomData,
+                            units: $crate::lib::marker::PhantomData,
+                            value: self.value.cbrt(),
+                        }
+                    }
+
+                    /// Fused multiply-add. Computes `(self * a) + b` with only one rounding error.
+                    /// This produces a more accurate result with better performance than a separate
+                    /// multiplication operation followed by an add.
+                    ///
+                    /// ## Generic Parameters
+                    /// * `Da`: Dimension for parameter `a`.
+                    /// * `Ua`: Base units for parameter `a`.
+                    /// * `Ub`: Base units for parameter `b`.
+                    #[inline(always)]
+                    pub fn mul_add<Da, Ua, Ub>(
+                        self,
+                        a: Quantity<Da, Ua, V>,
+                        b: Quantity<$quantities<$($crate::typenum::Sum<D::$symbol, Da::$symbol>),+>, Ub, V>,
+                    ) -> Quantity<$quantities<$($crate::typenum::Sum<D::$symbol, Da::$symbol>),+>, U, V>
+                    where
+                        $(D::$symbol: $crate::lib::ops::Add<Da::$symbol>,
+                        <D::$symbol as $crate::lib::ops::Add<Da::$symbol>>::Output: $crate::typenum::Integer,)+
+                        D::Kind: $crate::marker::Mul,
+                        Da: Dimension + ?Sized,
+                        Da::Kind: $crate::marker::Mul,
+                        Ua: Units<V> + ?Sized,
+                        Ub: Units<V> + ?Sized,
+                    {
+                        #[allow(unused_imports)]
+                        use $crate::num_traits::MulAdd;
+
+                        // (self * a) + b
+                        Quantity {
+                            dimension: $crate::lib::marker::PhantomData,
+                            units: $crate::lib::marker::PhantomData,
+                            value: self.value.mul_add(a.value, b.value),
+                        }
+                    }
+
+                    /// Raises a quantity to an integer power.
+                    ///
+                    #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
+                    #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
+                    /// # use uom::si::f32::*;
+                    /// # use uom::si::length::meter;
+                    /// use uom::typenum::P2;
+                    ///
+                    /// let a: Area = Length::new::<meter>(3.0).powi(P2::new());
+                    /// ```
+                    ///
+                    /// ## Generic Parameters
+                    /// * `E`: `typenum::Integer` power.
+                    #[inline(always)]
+                    pub fn powi<E>(
+                        self, _e: E
+                    ) -> Quantity<$quantities<$($crate::typenum::Prod<D::$symbol, E>),+>, U, V>
+                    where
+                        $(D::$symbol: $crate::lib::ops::Mul<E>,
+                        <D::$symbol as $crate::lib::ops::Mul<E>>::Output: $crate::typenum::Integer,)+
+                        D::Kind: $crate::marker::Mul,
+                        E: $crate::typenum::Integer,
+                    {
+                        Quantity {
+                            dimension: $crate::lib::marker::PhantomData,
+                            units: $crate::lib::marker::PhantomData,
+                            value: self.value.powi(E::to_i32()),
+                        }
+                    }
+
+                    /// Takes the square root of a number. Returns `NAN` if `self` is a negative
+                    /// number.
+                    ///
+                    #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust")]
+                    #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
+                    /// # use uom::si::f32::*;
+                    /// # use uom::si::area::square_meter;
+                    /// let l: Length = Area::new::<square_meter>(4.0).sqrt();
+                    /// ```
+                    ///
+                    /// The input type must have dimensions divisible by two:
+                    ///
+                    #[cfg_attr(all(feature = "si", feature = "f32"), doc = " ```rust,compile_fail")]
+                    #[cfg_attr(not(all(feature = "si", feature = "f32")), doc = " ```rust,ignore")]
+                    /// # use uom::si::f32::*;
+                    /// # use uom::si::length::meter;
+                    /// // error[E0271]: type mismatch resolving ...
+                    /// let r = Length::new::<meter>(4.0).sqrt();
+                    /// ```
+                    #[inline(always)]
+                    pub fn sqrt(
+                        self
+                    ) -> Quantity<
+                        $quantities<$($crate::typenum::PartialQuot<D::$symbol, $crate::typenum::P2>),+>,
+                        U, V>
+                    where
+                        $(D::$symbol: $crate::typenum::PartialDiv<$crate::typenum::P2>,
+                        <D::$symbol as $crate::typenum::PartialDiv<$crate::typenum::P2>>::Output: $crate::typenum::Integer,)+
+                        D::Kind: $crate::marker::Div,
+                    {
+                        Quantity {
+                            dimension: $crate::lib::marker::PhantomData,
+                            units: $crate::lib::marker::PhantomData,
+                            value: self.value.sqrt(),
+                        }
+                    }}
+                }
+            }
+        }
+
         impl<D, U, V> $crate::lib::clone::Clone for Quantity<D, U, V>
         where
             D: Dimension + ?Sized,
diff --git a/src/tests/system.rs b/src/tests/system.rs
index 1355f80..459c9c6 100644
--- a/src/tests/system.rs
+++ b/src/tests/system.rs
@@ -577,14 +577,6 @@ mod fixed {
     }
 }
 
-// Somehow Complex<f32|f64> don't implement these methods, even though they
-// should, bc f32 / f64 implement Float and FloatCore:
-//  - https://docs.rs/num/latest/num/struct.Complex.html#impl-5
-//  - https://docs.rs/num/latest/num/struct.Complex.html#impl-4
-//
-// See discussion here:
-// https://github.com/iliekturtles/uom/pull/287#pullrequestreview-922005933
-/*
 mod complex {
     storage_types! {
         types: Complex;
@@ -595,7 +587,7 @@ mod complex {
 
         #[test]
         fn fp_categories() {
-            let float_categories = vec![
+            let float_categories = [
                 VV::neg_infinity(),
                 VV::min_value(),
                 VV::zero(),
@@ -603,12 +595,13 @@ mod complex {
                 VV::infinity(),
                 VV::nan(),
             ];
-            for re in float_categories.clone() {
+
+            for re in float_categories {
                 for im in float_categories {
                     let complex = V::new(re, im);
 
                     // Infinities
-                    if re.is_infinite() || im.is_infinite() {
+                    if complex.is_infinite() {
                         assert!(!Length::new::<meter>(complex).is_finite());
                         assert!(Length::new::<meter>(complex).is_infinite());
                         assert!(!Length::new::<meter>(complex).is_normal());
@@ -616,7 +609,7 @@ mod complex {
                     }
 
                     // NaNs
-                    if re.is_nan() || im.is_nan() {
+                    if complex.is_nan() {
                         assert!(!Length::new::<meter>(complex).is_finite());
                         assert!(!Length::new::<meter>(complex).is_infinite());
                         assert!(!Length::new::<meter>(complex).is_normal());
@@ -624,8 +617,7 @@ mod complex {
                     }
 
                     // Finite and normal numbers
-                    if (!re.is_infinite() && !re.is_nan())
-                        && (!im.is_infinite() && !im.is_nan()) {
+                    if !complex.is_infinite() && !complex.is_nan() {
                         assert!(Length::new::<meter>(complex).is_finite());
                         assert!(!Length::new::<meter>(complex).is_infinite());
                         assert!(Length::new::<meter>(complex).is_normal());
@@ -638,26 +630,22 @@ mod complex {
         quickcheck! {
             #[allow(trivial_casts)]
             fn is_nan(v: A<V>) -> bool {
-                v.re.is_nan() || v.im.is_nan()
-                    == Length::new::<meter>(*v).is_nan()
+                v.is_nan() == Length::new::<meter>(*v).is_nan()
             }
 
             #[allow(trivial_casts)]
             fn is_infinite(v: A<V>) -> bool {
-                v.re.is_infinite() || v.im.is_infinite()
-                    == Length::new::<meter>(*v).is_infinite()
+                v.is_infinite() == Length::new::<meter>(*v).is_infinite()
             }
 
             #[allow(trivial_casts)]
             fn is_finite(v: A<V>) -> bool {
-                v.re.is_finite() && v.im.is_finite()
-                    == Length::new::<meter>(*v).is_finite()
+                v.is_finite() == Length::new::<meter>(*v).is_finite()
             }
 
             #[allow(trivial_casts)]
             fn is_normal(v: A<V>) -> bool {
-                v.re.is_normal() && v.im.is_normal()
-                    == Length::new::<meter>(*v).is_normal()
+                v.is_normal() == Length::new::<meter>(*v).is_normal()
             }
 
             #[cfg(feature = "std")]
@@ -672,16 +660,12 @@ mod complex {
                 Test::eq(&v.cbrt(), &l.value)
             }
 
-            #[cfg(feature = "std")]
-            #[allow(trivial_casts)]
-            fn hypot(l: A<V>, r: A<V>) -> bool {
-                Test::eq(&Length::new::<meter>(l.hypot(*r)),
-                    &Length::new::<meter>(*l).hypot(Length::new::<meter>(*r)))
-            }
-
             #[cfg(feature = "std")]
             #[allow(trivial_casts)]
             fn mul_add(s: A<V>, a: A<V>, b: A<V>) -> bool {
+                #[allow(unused_imports)]
+                use num_traits::MulAdd;
+
                 let r: Quantity<Q<P2, Z0, Z0>, U<V>, V> = Length::new::<meter>(*s).mul_add(
                     Length::new::<meter>(*a),
                     Quantity::<Q<P2, Z0, Z0>, U<V>, V> {
@@ -693,17 +677,6 @@ mod complex {
                 Test::eq(&s.mul_add(*a, *b), &r.value)
             }
 
-            #[allow(trivial_casts)]
-            fn recip(v: A<V>) -> bool {
-                let a: Quantity<Q<N1, Z0, Z0>, U<V>, V> = Quantity::<Q<P1, Z0, Z0>, U<V>, V> {
-                    dimension: PhantomData,
-                    units: PhantomData,
-                    value: *v,
-                }.recip();
-
-                Test::eq(&v.recip(), &a.value)
-            }
-
             #[cfg(feature = "std")]
             #[allow(trivial_casts)]
             fn powi(v: A<V>) -> bool {
@@ -724,7 +697,6 @@ mod complex {
         }
     }
 }
-*/
 
 mod non_complex {
     storage_types! {
diff --git a/src/unit.rs b/src/unit.rs
index c83f5c9..cf86fc7 100644
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -255,15 +255,16 @@ macro_rules! unit {
                 type T = VV;
 
                 #[inline(always)]
-                #[allow(trivial_numeric_casts)]
+                #[allow(clippy::inconsistent_digit_grouping)]
                 fn coefficient() -> Self::T {
-                    unit!(@coefficient $($conversion),+) as Self::T
+                    unit!(@coefficient $($conversion),+)
                 }
 
                 #[inline(always)]
-                #[allow(unused_variables, trivial_numeric_casts)]
+                #[allow(unused_variables)]
+                #[allow(clippy::inconsistent_digit_grouping)]
                 fn constant(op: $crate::ConstantOp) -> Self::T {
-                    unit!(@constant op $($conversion),+) as Self::T
+                    unit!(@constant op $($conversion),+)
                 }
             }
 
```

TobTobXX:
I still didn't really get why it works in the `float` block now, but it does. At least it compiles, still get failures everywhere, but that's one step closer.

iliekturtles:
> I still didn't really get why it works in the `float` block now, but it does. At least it compiles, still get failures everywhere, but that's one step closer.

The issue is that the original code implemented `is_nan` and all of the other floating point methods in an `impl<..., V> Quantity<..., V>` block `where V: Float`. `Complex<T>` doesn't implement `Float` so this original method didn't apply. `Complex<T>` *does* implement the methods outside the trait however. The solution is to remove the original method definitions in `uom` from the generic implementation and explicitly implement for each underlying storage type. e.g. `impl<..., f32> Quantity<..., f32>`, `impl<..., f64> Quantity<..., f64>`, `impl<..., Complex32> Quantity<..., Complex32>`, ... Manual monomorphization essentially.

I also reviewed the test failures. The issue is that `Complex<T>`'s div (and possibly `mul`?) implementation overflows with smaller numbers than regular `f32`/`f64` would. See [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=use%20num%3A%3Acomplex%3A%3AComplex32%3B%0A%0Afn%20main()%20%7B%0A%20%20%20%20let%20af%20%3D%201_f32%3B%0A%20%20%20%20let%20bf%20%3D%209.999999e-25_f32%3B%0A%20%20%20%20%2F%2Flet%20bf%20%3D%201e-24_f32%3B%0A%20%20%20%20let%20cf%20%3D%20af%20%2F%20bf%3B%0A%20%20%20%20let%20a%20%3D%20Complex32%3A%3Anew(af%2C%200E0)%3B%0A%20%20%20%20let%20b%20%3D%20Complex32%3A%3Anew(bf%2C%200E0)%3B%0A%20%20%20%20let%20c%20%3D%20a%20%2F%20b%3B%0A%20%20%20%20%2F%2Flet%20c%20%3D%20a.fdiv(b)%3B%0A%20%20%20%20%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20af)%3B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20bf)%3B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20cf)%3B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20a)%3B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20b)%3B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20c)%3B%0A%7D%0A) example. If you run you can see that `cf` is `1.0000001e24` while `c` is `inf`. Replacing the regular div call with `fdiv` gives the same results.

Other underlying storage types have similar issues (#44) so I wont block inclusion of complex. . If you believe there are no other open issues, will you squash your commits into a single commit (or a few that each make specific changes). I'll do a final review and believe we can accept.

TobTobXX:
> The issue is ...

Oh, I see.

> The issue is that Complex<T>'s div implementation overflows with smaller numbers than regular f32/f64 would.

> Other underlying storage types have similar issues (https://github.com/iliekturtles/uom/issues/44) so I wont block inclusion of complex.

Oh, yeah, would've never found that. Thank you! Well then, I guess we'll include them. Should I remove/comment out the tests to make the pipeline succeed?

Thank you a lot for your support on this PR!

iliekturtles:
Merged! Thank you so much for the PR and for your patience as I found time to work through the review. No need to comment out the tests as they aren't run by default. Currently I'm only requiring full test coverage for `f32`/`f64` and will enable the tests as precision issues are fixed for other types.

I'll likely look to do a new release to crates.io in the next couple weeks.

PR #291, opened 2022-03-23
Implement and test `exp` for Ratio

### LINKED ISSUE #290: Suport .exp() for Ratio
Is there any reason not to support `.exp()` for `Ratio`?

If not, I'd be happy to submit a PR.

jacg:
Close #290.

Please have a look if this is up to standard. If so, I'd be happy to add `exp2`, `ln`, `log`,  `log2` and `log10` in a similar fashion.

### DIFF
--- a/src/si/ratio.rs
+++ b/src/si/ratio.rs
@@ -32,13 +32,14 @@ quantity! {
     }
 }
 
-/// Implementation of various stdlib inverse trigonometric functions
+/// Implementation of various stdlib functions.
 #[cfg(feature = "std")]
 impl<U, V> Ratio<U, V>
 where
     U: crate::si::Units<V> + ?Sized,
     V: crate::num::Float + crate::Conversion<V>,
     radian: crate::Conversion<V, T = V::T>,
+    ratio: crate::Conversion<V, T = V::T>,
 {
     /// Computes the value of the inverse cosine of the ratio.
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


src/si/ratio.rs
```diff
@@ -77,6 +77,22 @@ where
     }
 }
 
+/// Implementation of various stdlib exponentiation and logarithm functions
+#[cfg(feature = "std")]
+impl<U, V> Ratio<U, V>
```
iliekturtles, line 82: Go ahead and merge this impl block with the trig block and update the comments. Same for the tests below. rename `mod inv_trig` to `mod float` and include all of the functions.

jacg: Done.

I have also included `exp_m1` and `ln_1p` which weren't in the original list.

`log` takes a second argument, besides `self`. I've given it the same type as `self` but I'm not sure whether we want to be more flexible. So I've added that one in a separate commit.

I've also added `[must_use = ...]` to the inverse-trig functions which were already there. Also in a separate commit, in case you don't want it. The `must_use` is there from the beginning for all the new functions that I added.

iliekturtles:
Things look good after a quick skim so I kicked off the test workflows. I'll review in full detail later today.

For the `[must_use = ...]` attributes I do want to keep them and will add a new issue to review other existing methods.

src/si/ratio.rs
```diff
@@ -32,49 +32,117 @@ quantity! {
     }
 }
 
-/// Implementation of various stdlib inverse trigonometric functions
+/// Implementation of various stdlib functions
```
iliekturtles, line 35: One minor change and I'll merge!

```suggestion
/// Implementation of various stdlib functions.
```

jacg: Done, via rebase, fixing the commit that changed that line in the first place.

iliekturtles: Thanks! I kicked of the test workflows again and will merge the PR once they're complete.

Issue #293, opened 2022-03-25
Conversion between Equivalent quantities with different underlying types


jacg:
(Is opening issues here, appropriate for asking questions about usage of `uom`? Is there some better place for discussions or asking for help?)

In my `uom`-less code, I have something that amounts to:

```rust
let a: f32 = todo!();
let b = a.floor();
let c: usize = b as usize;
```

Let's say that `a`, `b` and `c` are all `Length`s. How should this code be made unit-aware? I start off with something like this:

```rust
use uom::si::f32::Length as Flength;
use uom::si::usize::Length as Ulength;

let a: Flength = todo!();
let b = a.floor();
let c: Ulength = ??? b ???;
```

`as` cannot work for non-primitives, and my attempts to use any of the `From` machinery result in very noisy compiler error messages which amount to

```
the trait `From <Quantity ... f32 ... f32 ...> is not implemented for `Quantity ... usize ... usize ...`
```

Going off on a tangent, do you have any words of wisdom regarding extracting the signal from the almost pure noise in such error messages?

jacg:
Of course, it is possible to do something like
```rust
use uom::si::length::meter as arbitrary_conversion_unit;

Ulength::new::<arbitrary_conversion_unit>(b.get::<arbitrary_conversion_unit>() as usize);
```
but I'm hoping that there's a less ad-hoc, higher-level approach available.

adamreichold:
Note that the fields of `Quantity` are public, so if have you two quantities in the same system of units where only the base type is different, you can do something like

```rust
pub fn as_discrete_length(value: FLength) -> ULength {
    Quantity {
        dimension: PhantomData,
        units: PhantomData,
        value: value.value as usize,
    }
}
```

and possibly a generic extension by using some trait like [`num_traits::cast::NumCast`](https://docs.rs/num-traits/latest/num_traits/cast/trait.NumCast.html) to abstract over the conversion of value.

iliekturtles:
Posting questions as issue is the best way right now. It gives us a public record for anyone who may have a similar question in the future, allows others like to respond (@adamreichold has been right on this), and is, in some ways, an indication that documentation could be improved. I also haven't taken the time to see about enabling GitHub discussions (feedback on this last part welcome).

Both of the solutions mentioned so far work short-term, but I think long term the right solution is a conversion trait that takes advantage of `num_traits::cast` to do the dirty work. `QuantityCast` perhaps:

```rust
pub trait QuantityCast: ... {
[...]

PR #294, opened 2022-03-25
RFC: Add Quantity::as_ and Quantity::cast for changing the underlying storage type.

adamreichold:
A first go at something helping with #293.

The restriction to `$units<V>` instead of `U: Units<V>` seems a bit weird, but it was the only way I saw to enforce that the same set of base units would be used for the conversion. (The more general version fails with trait bound errors as e.g. the `f32` base units do not implement `Units<i32>` and without generic associated types there is not to ask a `Units<V>` implementor for its equivalent `Units<W>` implementor AFAIU.)

Finally, while a I added minimal doc tests, I am not sure where actual unit tests should end up as this involves at least two storages by definition.


adamreichold:
> without generic associated types there is no way to ask a Units<V> implementor for its equivalent Units<W> implementor AFAIU

I think we could add a separate trait

```rust
trait ChangeStorage<Ul, Ur>: Units<Ul> {
    type Units: Unit<Ur>;
}

impl<Ul, Ur> ChangeStorage<Ul, Ur> for $units<Ul> {
    type Units = $units<Ur>;
}
```

but how much this would w.r.t. generality as it would still cover only the `$units<_>` case by default.

iliekturtles:
Thanks for the PR! I'll take a look this coming week and see if I can figure out anything with the traits.

iliekturtles:
I don't think using `$units` is needed/desired.

The following diff compiles so it works... right?

I went a slightly different way and defined a `QuantityCast` trait although I suppose no trait or even `From`/`Into` (too implicit?) could be used. This is a very rough implementation and essentially delegates the conversion down to `num::cast::NumCast` which in turn relies on `num::cast::ToPrimitive`.

I'll do some thinking about usability over the weekend as I think the ability to do type conversions is proven, it's just a matter of determining what the interface should be.

```diff
diff --git a/src/lib.rs b/src/lib.rs
index be430b4..32b73cf 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -280,6 +280,10 @@ pub mod num {
     #[cfg(feature = "bigint-support")]
     pub use num_rational::BigRational;
 
+    pub mod cast {
+        pub use num_traits::cast::*;
+    }
+
     #[cfg(any(feature = "rational-support", feature = "bigint-support"))]
     pub mod rational {
         pub use num_rational::*;
diff --git a/src/system.rs b/src/system.rs
index 6324553..e9aac7c 100644
--- a/src/system.rs
+++ b/src/system.rs
@@ -590,6 +590,34 @@ macro_rules! system {
         impl_ops!(Sub, sub, -, SubAssign, sub_assign, -=, Diff,
             Div, div, /, DivAssign, div_assign, /=, sub_div);
 
+        trait QuantityCast<D, U, V>: Sized
+        where
+            D: Dimension + ?Sized,
+            U: Units<V> + ?Sized,
+            V: $crate::num::Num + $crate::num::cast::ToPrimitive + $crate::Conversion<V>,
+        {
+            fn from(q: Quantity<D, U, V>) -> Option<Self>;
+        }
+
+        impl<D, Ur, Ul, Vr, Vl> QuantityCast<D, Ul, Vl> for Quantity<D, Ur, Vr>
+        where
+            D: Dimension + ?Sized,
+            Ur: Units<Vr> + ?Sized,
+            Ul: Units<Vl> + Units<Vr> + ?Sized,
+            Vr: $crate::num::Num + $crate::num::cast::NumCast + $crate::Conversion<Vr>,
+            Vl: $crate::num::Num + $crate::num::cast::ToPrimitive + $crate::Conversion<Vl>,
+        {
+            fn from(q: Quantity<D, Ul, Vl>) -> Option<Self>
+            {
+                <Vr as $crate::num::cast::NumCast>::from(q.value)
+                    .map(|v| Quantity {
+                        dimension: $crate::lib::marker::PhantomData,
+                        units: $crate::lib::marker::PhantomData,
+                        value: change_base::<D, Ul, Ur, Vr>(&v),
+                    })
+            }
+        }
+
         impl<D, U, V> Quantity<D, U, V>
         where
             D: Dimension + ?Sized,
```

adamreichold:
The trait bound which did not work out for me was

```rust
Ul: Units<Vl> + Units<Vr> + ?Sized,
```

as e.g. `SI<f32>` would only implement `Units<f32>` but not say `Units<i32>`.

dzmitry-lahoda:
casts are good, for example https://github.com/iliekturtles/uom/pull/529 i would like casts because inputs are 64 bit, but I lift system to 128 bit as needed for aggregations

Issue #295, opened 2022-03-29
Specifying ad-hoc Quantities and functions generic over Quantities


jacg:
Consider this toy function
```rust
fn invert(x: Length) -> PerLength {
    1.0 / x
}
```

1. Is there a convenient way of defining ad-hoc `Quantity`s such as `PerLength`? Can I do better than this
    ```rust
    use uom::si::{ISQ, SI, Quantity};
    use uom::typenum::{Z0, N1};

    type PerLength = Quantity<ISQ<N1, Z0, Z0, Z0, Z0, Z0, Z0>, SI<f32>, f32>;
    ```
    ?
2. How can this function be made generic over the input (and output) Quantity?
    Naively, I would like to write
    ```rust
    fn invert<T>(x: T) -> Per<T> {
        1.0 / x
    }
    ```
    + `Per<T>` can be expressed with `typenum` gymnastics, such as used in [the return type of `sqrt`](https://docs.rs/uom/latest/uom/si/struct.Quantity.html#method.sqrt).
    + `fn invert<T>(x: T)` will probably look more like `fn invert<D, U, V>(x: Quantity<D, U, V>)`, but when I try to add the necessary bounds on `D`, `U` and `V` I always end up getting stuck.

jacg:
Another related problem arises in the context of #289:

> Do you see use for affine/vector quantities beyond length? Are there cases in your simulation where Point and Vector are not sufficient?

I have `Vector` and `Point` types, each containing 3 `Length`s. Parts of the code are expressed in terms of normalized units, relative to some unit cell. Those parts require `Vector` and `Point` types containing 3 `Ratio`s each.

I kinda anticipated this, and wanted to make `Vector` and `Point` generic over the contained `Quantity`, but got lost in the implementation, and went with the hard-wired `Length` to start with. But now the motivation for the generic versions is much stronger.

Ideally I'd like to have `Vector<T>` and `Point<T>` such that

+ `Vector<Length> + Vector<Length> -> Vector<Length>`
+ `Vector<Length> / Vector<Length> -> Vector<Ratio>`
+ `Point<Length> + Vector<Length> -> Point<Length>`
+ `Point<Ratio> + Vector<Ratio> -> Point<Ratio>`

Is it possible to implement these in a generic way?

iliekturtles:
[...]

PR #298, opened 2022-04-01
Make `Units` trait alias created by `$quantities!` macro `pub`.

iliekturtles:
Part of #295 where the need to access the `Units` trait for a
user-defined system of units was uncovered.

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -1518,10 +1518,11 @@ macro_rules! system {
         /// * `$system`: Path to the module where the [`system!`](macro.system.html) macro was run
         ///   (e.g. `uom::si`).
         /// * `$V`: Underlying value storage type (e.g. `f32`).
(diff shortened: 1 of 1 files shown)


src/system.rs
```diff
@@ -1628,7 +1629,13 @@ macro_rules! system {
     ) => {
         use $path as __system;
 
-        type Units = dyn __system::Units<$V, $($name = __system::$name::$U,)+>;
+        /// Marker trait to identify a [system of units][units] based on the set of given [base
```
adamreichold, line 1632: This is not a trait, is it? Rather a placeholder for an implementor of the trait with concrete base units filled in?

iliekturtles: Updated the wording slightly. This make more sense?

```rust
        /// `Units` trait type alias to identify a [system of units][units] based on the set of
        /// given [base units][base] of a [system of quantities][quantities].
```

PR #307, opened 2022-05-25
[WIP] Operations for Quantity references

### LINKED ISSUE #38: Implement operations for Quantity references
`impl Op<Quantity<...>> for Quantity<..>` is already complete. Add the following reference implementations:

- [ ] `impl<'a, 'b> Op<&'a Quantity<...>> for &'b Quantity<..>`
- [ ] `impl<'a> Op<&'a Quantity<...>> for Quantity<..>`
- [ ] `impl<'a> Op<Quantity<...>> for &'a Quantity<..>`

jacg:
Fix #38.

Prototype implementations of 

1. `&q +/- q`
2. `q +/- &q`
3. `&q +/- &q`

Could you check that I'm barking up the right tree, before I try to complete the rest?


jacg:
Apologies for the long delay, my time to work on this is sporadic.

I agree, the code duplication is quite horrendous. When I try to make sense of it, I get lost and my eyes glaze over.

I've taken the liberty of summarizing your comment, broken down into enumerated questions:

1. Can `autoconvert!` duplication be merged?
2. Does introducing references lead to quadruple code repetition?
3. Do `Copy` vs non`Copy` underlying storage types introduce further doubling?
4. Should we forward to same operator?

Faced with all of these at once, I don't know where to begin. Do you think we could identify one of them as a candidate on which to start making some progress?

iliekturtles:
I've been doing some investigation and believe we can eliminate all of the duplication, include the existing duplication for `autoconvert`. Since [1.13.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#language-48) macros are allowed in the type position. This wasn't the case when the original stdlib code was written and I'm guessing no one went back to reduce duplication after that.

I wrote out some example code in the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=49c466bf35da9359480ecbe6b26f2b71) showing how type parameters and references can be inserted and I believe we can put together a recursive macro as follows:

 1. Outer call injects which units type parameters should be used.
```rust
#[cfg(not(feature = "autoconvert"))]
impl_ops!(U);

#[cfg(feature = "autoconvert")]
impl_ops!(Ul, Ur);
```

 2. First branch injects the by-val/by-ref combinations. This could possibly be merged with the outer call since there really isn't much duplication. This may also *need* to be merged with the previous step because `$()` repetitions that aren't nested in the pattern can't be nested in the body. This would essentially mean hand-writing the 8 combinations (e.g. `((U), V, V), ((Ul, Ur), V, V), ((U), V, &V), ((Ul, Ur), V, &V), ...`).
```rust
 macrol_rules! impl_ops {
    ($(U:ty),+) => {
        impl_ops!(@types ($($U),+), ((V, V), (V, &V), (&V, V), (&V, &V));
    };
}

```

 3. Second branch injects the traits to implement. This is basically the original code with the U and V information added in.
```rust
 macrol_rules! impl_ops {
    (@types ($($U:ty),+), ($(($Vl:ident, $Vr:ident)),+)) => {
        impl_ops!(@impl ($($U),+), ($(($Vl, $Vr)),+), Add, add, +, ...);
        impl_ops!(@impl ($($U),+), ($(($Vl, $Vr)),+), Sub, sub, -, ...);
    };
}
```

 4. Innermost branch has the actual implementations. Note how usages of `U`, `Ur`, `Ul`, and `V` are replaced by macro calls to transform them into the appropriate actual type.
```rust
 macrol_rules! impl_ops {
    (@types ($($U:ty),+), ($(($Vl:ident, $Vr:ident)),+), ...) => {

        impl<D, $(U,)+, V> $AddMul<Quantity<D, impl_ops!(@Ur $($U),+), impl_ops!(@Vr $($Vl, $Vr),+)> for ...
    };
}
```

 5. Finally, extra match arms without repetitions can do the actual transformation.
```rust
 macrol_rules! impl_ops {
    (@Ul $U:ty) => { $U };
    (@Ul $Ul:ty, $Ur:ty) => { $Ul };
    (@Ur $U:ty) => { $U };
    (@Ur $Ul:ty, $Ur:ty) => { $Ur };
    (@Vl $Vl:ident, $Vr:ident) => { $Vl };
    (@Vr $Vl:ident, $Vr:ident) => { $Vr };
}
```

Let me know how you want to proceed and how I can assist. You might want to start out with a simple playground or test project and use `cargo expand` to view the output.

jacg:
So you can think of the whole recursive macro invocation graph as a tree, with macro invocations at the nodes, and actual code (`impl`s) as the leaves.

I think that the way forward is to start at the bottom node-level, and manually invoke the macro to generate a bunch of related leaves. Then repeat it for some siblings of the node, to make sure the pattern is consistent. And then try to go up one level, and replace the manual invocations at the lowest node level, with ones generated by a manual invocation at the higher level. And so on, until the whole lot is replaced with one big top-level invocation.

This should make it possible to gradually replace (and fill gaps in) the existing code while keeping the whole lot compiling throughout.

Does this sound reasonable? Can you see any obvious flaws in this strategy? Can you think of a good candidate for a first bottom node to try?

iliekturtles:
That sounds like a good way to start to me. I would just make sure that you include all the parameters that will eventually be needed (`$($U)`, `$Vl`, `$Vr`).

```rust
macro_rules! impl_ops {
    (($($U:ident),+), $Vl:ty, $Vr:ty, $AddSubTrait:ident, ...) => {
        impl<D, $($U,)+ V> $crate::lib::ops::$AddSubTrait<Quantity<D, impl_ops!(@Ur $($U),+), $Vr>> for Quantity<D, impl_ops!(@Ul $($U),+), $Vl> {
            // ...
        })+
    }
```

jacg:
Could you help me make sense of the existing implementations?

Looking at the original code, I see the following nine `impl`s:

|   | lines                                                                                                                | ops  |     | autoconv              |         |         |   | `AddSub`           | `MulDiv`           |                    |
|---|----------------------------------------------------------------------------------------------------------------------|------|-----|-----------------------|---------|---------|---|--------------------|--------------------|--------------------|
| 1 | [394-414](https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L394-L414) | `+-` |     | :heavy_check_mark:    | `Ul,Ur` | `D`     |   | :heavy_check_mark: | :x:                |                    |
| 2 | [416-434](https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L416-L434) | `+-` |     | :x:                   | `U`     | `D`     |   | :heavy_check_mark: | :x:                |                    |
| 3 | [436-451](https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L436-L451) | `+-` | `=` | :heavy_check_mark:    | `Ul,Ur` | `D`     |   | :heavy_check_mark: | :x:                |                    |
| 4 | [453-466](https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L453-L466) | `+-` | `=` | :x:                   |         | `D`     |   | :heavy_check_mark: | :x:                |                    |
|   |                                                                                                                      |      |     |                       |         |         |   |                    |                    |                    |
| 5 | [468-494](https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L468-L494) | `*/` |     | :heavy_check_mark:    | `Ul,Ur` | `Dl,Dr` |   | :heavy_check_mark: | :heavy_check_mark: |                    |
| 6 | [496-521](https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L496-L521) | `*/` |     | :x:                   | `U`     | `Dl,Dr` |   | :heavy_check_mark: | :heavy_check_mark: |                    |
| 7 | [523-540](https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L523-L540) | `*/` |     | :heavy_check_mark::x: | `U`     | `D`     |   | :x:                | :heavy_check_mark: | dimensionless RHS? |
| 8 | [542-554](https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L542-L554) | `*/` | `=` | :heavy_check_mark::x: | `U`     | `D`     |   | :x:                | :heavy_check_mark: | dimensionless RHS? |
|   |                                                                                                                      |      |     |                       |         |         |   |                    |                    |                    |
| 9 | [556-587](https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L556-L587) | `*/` |     | :heavy_check_mark::x: | `U`     | `D`     |   | :heavy_check_mark: | :heavy_check_mark: | in a `mod`, `Z0`   |
    
The first four only use add-sub related names, and cover the four possible combinations of assignment-autoconversion. This all makes sense.

The next four are trickier: 

+ Even though they implement mul-div, they use add-sub names for type-level arithmetic on the dimensions.
+ The assign versions cannot work in general, only when the RHS is a `Ratio` or some plain float.
+ The two that don't use add-sub names (7 & 8) must be covering the RHS-is-dimensionless case (once with `=`, once without), but I don't see how that constraint is expressed. There is a `typenum::Z0` in no. 9, but not here.

Then there is a ninth `impl` inside a module. What's that? I see that it's the only one which has a `typenum::Z0` constraint.

I would have expected autoconversion only to make sense in the presence of `Ul, Ur`, but 7, 8 and 9 are counterexamples.

iliekturtles:
* When you add or subtract quantities the dimensions stay the same. e.g. `length + length = length`. To write in terms of quantity `Quantity<D, U, V> + Quantity<D, U, V> = Quantity<D, U, V>`. The `D` parameter does not change. When you multiply or divide the dimensions change. e.g. `length / time = velocity`. Written in terms of quantity `Quantity<Dl, U, V> / Quantity<Dr, U, V> = Quantity<Doutput, U, V>`. To calculate the this change for multiplication or division the powers of the [base quantities](https://en.wikipedia.org/wiki/International_System_of_Quantities#Base_quantities) are added or subtracted: length (`L¹T⁰`) / time (`L⁰T¹`) = length (`L¹T⁰`) * frequency (`L⁰T⁻¹`) = velocity (`L¹⁺⁰T⁰⁺⁻¹`).
 * Why can't the assign versions work in general? Because they only have a single `U` parameter? Instead of using the enumeration of `U` parameters in the macro, just use `Ul` (Or `Ur` for 9).
 * 7 and 8 implement multiply and divide when RHS is a scalar (plain number, e.g. `f32`) and maintain the dimensions. e.g. `Quantity<D, U, V> */ f32 = Quantity<D, U, V>`. 9 implements multiply and divide when LHS is a scalar and RHS is a quantity. e.g. `f32 */ Quantity<D, U, V>`. You can think of this impl as if it were `f32.into_ratio_quantity() */ Quantity<D, U, V>` and then look at the definition for 5 replacing `Dl` with a dimension where all base quantities are `Z0` and `Ul` with `Ur`. For multiply this will maintain the dimensions and scale the value. For divide this will essentially invert the quantity. e.g. `1 / time = frequency`.

   The `mod` is used because the `impl<D, U>` doesn't introduce a `V` parameter (would be an orphan rule [violation](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d0c0db6f257ceeffbfe49051e8ad2e54)) so the `storage_types!` macro is used to define a `V` type alias that can be used. `storage_types!` also introduces new mod names so the outer mod effectively hides them so they don't cause conflicts with later uses of `storage_types!`.

jacg:
I have realistic hope of being able to spend some time on this, this week.

I've been wondering about what you wrote here:

> I wrote out some example code in the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=49c466bf35da9359480ecbe6b26f2b71) showing how type parameters and references can be inserted

I understand how these macro tricks work, but I don't understand why we need them.  What does the version on the first line here
```rust
impl<X, Y> T<x!(@R X, Y)> for S1<x!(@P X, Y)> {}
impl<X, Y> T<         Y > for S2<     &X    > {}
```
([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5d16606bce9b58558c0c9c643af7fdc2)) give us, that the version on the second line does not (apart from noise and obfuscation)?

Perhaps I'm missing something about how this information will propagate from the top of the macro call, but out of that context it seems that the first line uses `x!(@R X, Y)` where it could simply say `Y`, and `x!<@P X, Y>` as a long-winded and cryptic way of saying `&X`.

Also, I've been wondering whether `impl_ops` should be a single recursive macro with lots of match arms playing a variety of different roles corresponding to different levels of the invocation tree, or whether it wouldn't be better to have different macros for each level, or maybe even have different macros for different sub-trees at the same level.

jacg:
Do you have any words of wisdom on how to test this?

I've found https://github.com/iliekturtles/uom/blob/d55d64ac62a7f6d56a907f8f2630fa50701b55b8/src/tests/quantity.rs#L54-L129 and https://github.com/iliekturtles/uom/blob/d55d64ac62a7f6d56a907f8f2630fa50701b55b8/src/tests/system.rs#L135-L188 and some temperature-specifc ones here https://github.com/iliekturtles/uom/blob/d55d64ac62a7f6d56a907f8f2630fa50701b55b8/src/si/thermodynamic_temperature.rs#L176-L235

I have used Haskell's Quickcheck, *many* moons ago and I use [Proptest](https://crates.io/crates/proptest) in Rust, but I've never used Rust's Quickcheck, so these tests are a bit cryptic to me.

I'm tempted to write some very simple example-based tests, to try to make it blatantly obvious exactly what is being tested, and what isn't in each test. Even if we don't keep them, they should serve as a useful crutch during development.

jacg:
IIUC

+ the `autoconvert` feature is enabled by `cargo test`,
+ the purpose of `autoconvert` is to enable mixing of base units in the arithmetic operations.

Consequently, I would expect variations on this theme to compile
```rust
    #[test]
    fn hmm() {
        use crate::si::time::second;
        let base_f32 = crate::si::f32::Time::new::<second>(1.0);
        let base_f64 = crate::si::f64::Time::new::<second>(1.0);
        base_f32 + base_f64;
    }
```
but I always get errors along these lines
```
error[E0308]: mismatched types
  --> src/tests/impl_ops.rs:55:20
   |
55 |         base_f32 + base_f64;
   |                    ^^^^^^^^ expected `f32`, found `f64`
   |
[...]
```
What am I doing wrong?

iliekturtles:
Responding to your first comment:

> I've been wondering about what you wrote here:
> 
> > I wrote out some example code in the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=49c466bf35da9359480ecbe6b26f2b71) showing how type parameters and references can be inserted
> 
> I understand how these macro tricks work, but I don't understand why we need them. What does the version on the first line here
> 
> ```rust
> impl<X, Y> T<x!(@R X, Y)> for S1<x!(@P X, Y)> {}
> impl<X, Y> T<         Y > for S2<     &X    > {}
> ```
> 
> ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5d16606bce9b58558c0c9c643af7fdc2)) give us, that the version on the second line does not (apart from noise and obfuscation)?

The intention was to parameterize if the reference on the type.  When I was roughly writing out the example match arms in https://github.com/iliekturtles/uom/pull/307#issuecomment-1172917574 I had it in my mind that the reference would be applied to the `V` parameter. This is not the case! The possible reference gets applied to `Quantity`. e.g. `impl Add for Quantity<...>` and `impl Add for &Quantity<...>`.

Here is another [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ca729f841a05e82f2012c74819e3def5) that I hope better illustrates what I believe needs to be done. To rephrase some of the earlier example code the parameterization should probably look something like this with the different `U`/`Ul`/`Ur` and reference combinations: `((U), , ), ((Ul, Ur), , ), ((U), , &), ((Ul, Ur), , &), ((U), &, ), ((Ul, Ur), &, ), ((U), &, &), ((Ul, Ur), &, &))`

Let me know if this part makes sense. I'll respond to the rest of the questions later today or tomorrow.

iliekturtles:
> Also, I've been wondering whether `impl_ops` should be a single recursive macro with lots of match arms playing a variety of different roles corresponding to different levels of the invocation tree, or whether it wouldn't be better to have different macros for each level, or maybe even have different macros for different sub-trees at the same level.

I'm leaning towards a single macro with multiple "private" match arms.

> Do you have any words of wisdom on how to test this?

I haven't thought about the best way to organize tests for all of the different combination of ops impls. For reference the first block you linked is for when units with different base units are operated on. The second is the main set of tests for the ops. The third is for temperature specific explicit implementations (certain ops are not implemented for thermodynamic temperature because of it's kind).

`quickcheck!` is essentially just calling the inner functions with random values in the parameters.

> the purpose of autoconvert is to enable mixing of base units in the arithmetic operations.
> but I always get errors along these lines

`autoconvert` allows for quantities with different base units (e.g. `length` stored as `meter` vs. `length` stored as `kilometer`) _not_ different underlying storage types (`f32`, `f64`).

jacg:
I'm trying to write an exhaustive set of simple-to-understand tests, covering all the things that should work. When I have all of these written out by hand, I'll try to write macros to generate them. These should have much in common, structurally, with the macros `impl_ops` macro that is to be used to generate the implementations. This should give me better insight into how `impl_ops` will need to be implemented, and some practice in actually implementing them, in a simpler context. So I'll come back to your macrology suggestions, once I've done that.

For now

```
<Quantity> *= <float>
```
works, but

```
<Quantity> *= <Ratio>
```
fails to compile. [And similarly for `/=`.]

Is the latter not supposed to work, or am I making a mistake somewhere?

iliekturtles:
Good catch. `MulAssign`/`DivAssign` are currently only implemented for the underlying storage type. 

https://github.com/iliekturtles/uom/blob/10d9679193775aae04987338a31014524bd5674d/src/system.rs#L542-L554

I don't see why it couldn't explicitly be implemented for `Ratio` (`Quantity<DN<Z0>, ...>`). It's possible there are trait bound issues, but I don't remember if I even tried to do the implementation.

```rust
impl<D, Ul, Ur, V> $crate::lib::ops::$MulDivAssignTrait<Quantity<DN<Z0>, Ur, V> for Quantity<D, Ul, V>
where ...
{
    ...
}
```

https://github.com/iliekturtles/uom/blob/10d9679193775aae04987338a31014524bd5674d/src/system.rs#L274-L276

jacg:
I've pushed a commit which adds some very concise tests, whose purpose is to summarize what operations currently work.

Nothing fancy, no exhaustive property-based testing, just a single hand-picked example for each case, and some macros to maximize the signal-to-noise ratio in the test collection, so that it can act as a quick overview for a human reader. The layout is aggressively aligned to make parsing and comparison by *human* eye as easy as possible, so it will make any rustfmt checks have a fit, but those are irrelevant for the time being.

Could you glance over https://github.com/iliekturtles/uom/blob/4de23e8e6cb173b06abe70bd2d9559713653a923/src/tests/impl_ops.rs#L95-L124
to check that I have tested everything that you expect to work at present, that the test names correctly reflect what is being tested, and that the numbers in the comments at the end of each line match the relevant implementations listed in https://github.com/iliekturtles/uom/pull/307#issuecomment-1186208970.

As noted earlier, the assignment operators with `Ratio` on the RHS do not work, so those test are present, but commented out.

Would a small PR for the currently missing `MulDivAssign Ratio` case (if it can be made to work, haven't looked into it yet) make sense, *before* tackling the reference operand implementations in this PR?

jacg:
I have added (in the latest commit, at present)
```rust
impl<D, Ul, Ur, V> $crate::lib::ops::$MulDivAssignTrait<Quantity<DimensionOne, Ur, V>> for Quantity<D, Ul, V>
where ...
```
but the relevant tests continue to fail to compile with exactly the same error message as before:
```
126 |         test!(mul_assign_ratio    [    m(4.) ] *= ratio(2.),        m(8.)); //  ERR      c.f. AAA above
    |                                                   ^^^^^^^^^ expected `f32`, found struct `si::Quantity`
```
Any idea why it's not picking up this new implementation as appropriate for the requested operation?

jacg:
> A separate PR to add `MulDivAssign` for `Ratio` is a good idea.

Done in #364.

PR #309, opened 2022-06-08
fix #283: add possibility to include num-traits/libm for sin and cos …

### LINKED ISSUE #283: enable cos and sin on `no_std` builds
Currently `sin` and `cos` functions are disabled on `no_std`. These are common functions which are easily available through `libm`, so it makes sense to expose them.

wucke13:
…in no_std builds


src/features.rs
```diff
@@ -81,11 +81,11 @@ macro_rules! si {
...
 }
 
-/// Expands the given block of code when `uom` is compiled with the `std` feature.
+/// Expands the given block of code when `uom` is compiled with either the `std` or the `libm` feature.
 #[doc(hidden)]
 #[macro_export]
-#[cfg(feature = "std")]
-macro_rules! std {
+#[cfg(any(feature = "std", feature = "libm"))]
+macro_rules! float {
```
iliekturtles, line 88: Revert these changes and instead add two new macros, `std_libm`, that can be used to conditionally expand the nested code when either of the features is enabled. See `autoconvert_test` as an example.

iliekturtles (commented):
Thanks for the PR! I kicked off the build. I expect we'll see a couple errors come up from that.

src/features.rs
wucke13: I did so in the first place, however, I think this doesn't make sense (why I unified the two). There are only two places allover the `uom` src where the `std!` macro was used (both in `src/system.rs`). And in both places, the usage of `std!` is actually only about using using float related operations IIRC. So it makes sense to not use the `std!` macro at all, but always the `std_libm!` feature.

Based on that assumption, it also makes sense to remove the `std!` completely (since it is unused) and I think ~~`float!` is more accurate than `std_libm!`, but that's up to debate (most descriptive of cause would be `std_or_libm!`)~~ `float` creates name collisions, so I used `std_or_libm` instead . 

Is this a convincing argument?
Do you want to keep the `std!`?

wucke13:
Fixed yet another bug, now it could work

wucke13:
Probably this is still not ready, I got some errors (even so the test suite runs through flawlesly:

```
error[E0599]: no method named `cbrt` found for type `f64` in the current scope
   --> /home/wucke13/.cargo/git/checkouts/uom-c45462707dc2a646/f82bc45/src/system.rs:844:47
    |
844 |   ...                   value: self.value.cbrt(),
    |                                           ^^^^ method not found in `f64`
    |
   ::: /home/wucke13/.cargo/git/checkouts/uom-c45462707dc2a646/f82bc45/src/si/mod.rs:10:1
    |
10  | / system! {
11  | |     /// [International System of Quantities](https://jcgm.bipm.org/vim/en/1.6.html) (ISQ).
12  | |     ///
13  | |     /// ## Generic Parameters
...   |
104 | |     }
105 | | }
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following traits are implemented but not in scope; perhaps add a `use` for one of them:
            candidate #1: `use crate::num_traits::Float;`
            candidate #2: `use crate::num_traits::real::Real;`
    = note: this error originates in the macro `system` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: no method named `powi` found for type `f64` in the current scope
   --> /home/wucke13/.cargo/git/checkouts/uom-c45462707dc2a646/f82bc45/src/system.rs:910:47
    |
910 |   ...                   value: self.value.powi(E::to_i32()),
    |                                           ^^^^ method not found in `f64`
    |
   ::: /home/wucke13/.cargo/git/checkouts/uom-c45462707dc2a646/f82bc45/src/si/mod.rs:10:1
    |
10  | / system! {
11  | |     /// [International System of Quantities](https://jcgm.bipm.org/vim/en/1.6.html) (ISQ).
12  | |     ///
13  | |     /// ## Generic Parameters
...   |
104 | |     }
105 | | }
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following traits are implemented but not in scope; perhaps add a `use` for one of them:
            candidate #1: `use crate::ConversionFactor;`
            candidate #2: `use crate::num_traits::float::FloatCore;`
            candidate #3: `use crate::num_traits::Float;`
            candidate #4: `use crate::num_traits::real::Real;`
            candidate #5: `use crate::typenum::Pow;`
    = note: this error originates in the macro `system` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: no method named `sqrt` found for type `f64` in the current scope
   --> /home/wucke13/.cargo/git/checkouts/uom-c45462707dc2a646/f82bc45/src/system.rs:948:47
    |
948 |   ...                   value: self.value.sqrt(),
    |                                           ^^^^ method not found in `f64`
    |
   ::: /home/wucke13/.cargo/git/checkouts/uom-c45462707dc2a646/f82bc45/src/si/mod.rs:10:1
    |
10  | / system! {
11  | |     /// [International System of Quantities](https://jcgm.bipm.org/vim/en/1.6.html) (ISQ).
12  | |     ///
13  | |     /// ## Generic Parameters
...   |
104 | |     }
105 | | }
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following traits are implemented but not in scope; perhaps add a `use` for one of them:
            candidate #1: `use crate::num_traits::Float;`
            candidate #2: `use crate::num_traits::real::Real;`
    = note: this error originates in the macro `system` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0599`.
error: could not compile `uom` due to 3 previous errors
warning: build failed, waiting for other jobs to finish...
```

wucke13:
I got this from running cargo build from a project where I use `uom`, however, I spent half an hour now trying to recreate it and couldn't.

wucke13:
Anyway, I replace the three offending replaces of freestanding calls to the float numbers with explicit mentioning of the `num_traits::Float` trait.

wucke13:
@iliekturtles Ok, now I could use your help. Never worked with `Complex<_>`.

iliekturtles:
`Complex` doesn't `impl Float`, but it does implement most of `Float`'s methods. Changing to directly calling the function seems to be the trick. See the patch below that does this, adds a few missing `libm` feature checks, and uses `libm` in the CI tests.

```diff
diff --git a/.github/workflows/ci-full-test-suite.yml b/.github/workflows/ci-full-test-suite.yml
index 8d6c70f..2d70e0f 100644
--- a/.github/workflows/ci-full-test-suite.yml
+++ b/.github/workflows/ci-full-test-suite.yml
@@ -48,7 +48,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --verbose --no-default-features --features "autoconvert f32 si use_serde"
+          args: --verbose --no-default-features --features "autoconvert f32 si use_serde libm"
 
       - name: Test si with underlying storage types
         uses: actions-rs/cargo@v1
diff --git a/src/system.rs b/src/system.rs
index 05ac4ef..e68bc37 100644
--- a/src/system.rs
+++ b/src/system.rs
@@ -841,7 +841,7 @@ macro_rules! system {
                         Quantity {
                             dimension: $crate::lib::marker::PhantomData,
                             units: $crate::lib::marker::PhantomData,
-                            value: $crate::num_traits::Float::cbrt(self.value),
+                            value: self.value.cbrt(),
                         }
                     }
 
@@ -907,7 +907,7 @@ macro_rules! system {
                         Quantity {
                             dimension: $crate::lib::marker::PhantomData,
                             units: $crate::lib::marker::PhantomData,
-                            value: $crate::num_traits::Float::powi(self.value, E::to_i32()),
+                            value: self.value.powi(E::to_i32()),
                         }
                     }
 
@@ -945,7 +945,7 @@ macro_rules! system {
                         Quantity {
                             dimension: $crate::lib::marker::PhantomData,
                             units: $crate::lib::marker::PhantomData,
-                            value: $crate::num_traits::Float::sqrt(self.value),
+                            value: self.value.sqrt(),
                         }
                     }}
                 }
diff --git a/src/tests/quantity.rs b/src/tests/quantity.rs
index 00ceced..159d8ce 100644
--- a/src/tests/quantity.rs
+++ b/src/tests/quantity.rs
@@ -468,7 +468,7 @@ mod float {
             Test::assert_eq(&3.3.fract(), &m1.fract::<kilogram>().get::<kilogram>());
         }
 
-        #[cfg(feature = "std")]
+        #[cfg(any(feature = "std", feature = "libm"))]
         quickcheck! {
             #[allow(trivial_casts)]
             fn hypot_same(l: V, r: V) -> bool {
@@ -477,7 +477,7 @@ mod float {
             }
         }
 
-        #[cfg(all(feature = "std", feature = "autoconvert"))]
+        #[cfg(all(any(feature = "std", feature = "libm"), feature = "autoconvert"))]
         quickcheck! {
             #[allow(trivial_casts)]
             fn hypot_mixed(l: V, r: V) -> bool {
diff --git a/src/tests/system.rs b/src/tests/system.rs
index 1b5399f..46aa6dc 100644
--- a/src/tests/system.rs
+++ b/src/tests/system.rs
@@ -286,7 +286,7 @@ mod float {
                 v.classify() == Length::new::<meter>(*v).classify()
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(any(feature = "std", feature = "libm"))]
             #[allow(trivial_casts)]
             fn cbrt(v: A<V>) -> bool {
                 let l: Quantity<Q<P1, Z0, Z0>, U<V>, V> = Quantity::<Q<P3, Z0, Z0>, U<V>, V> {
@@ -298,7 +298,7 @@ mod float {
                 Test::eq(&v.cbrt(), &l.value)
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(any(feature = "std", feature = "libm"))]
             #[allow(trivial_casts)]
             fn hypot(l: A<V>, r: A<V>) -> bool {
                 Test::eq(&Length::new::<meter>(l.hypot(*r)),
@@ -315,7 +315,7 @@ mod float {
                 v.is_sign_negative() == Length::new::<meter>(*v).is_sign_negative()
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(any(feature = "std", feature = "libm"))]
             #[allow(trivial_casts)]
             fn mul_add(s: A<V>, a: A<V>, b: A<V>) -> bool {
                 let r: Quantity<Q<P2, Z0, Z0>, U<V>, V> = Length::new::<meter>(*s).mul_add(
@@ -340,13 +340,13 @@ mod float {
                 Test::eq(&v.recip(), &a.value)
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(any(feature = "std", feature = "libm"))]
             #[allow(trivial_casts)]
             fn powi(v: A<V>) -> bool {
                 Test::eq(&v.powi(3), &Length::new::<meter>(*v).powi(P3::new()).value)
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(any(feature = "std", feature = "libm"))]
             #[allow(trivial_casts)]
             fn sqrt(v: A<V>) -> TestResult {
                 if *v < V::zero() {
@@ -647,7 +647,7 @@ mod complex {
                 v.is_normal() == Length::new::<meter>(*v).is_normal()
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(any(feature = "std", feature = "libm"))]
             #[allow(trivial_casts)]
             fn cbrt(v: A<V>) -> bool {
                 let l: Quantity<Q<P1, Z0, Z0>, U<V>, V> = Quantity::<Q<P3, Z0, Z0>, U<V>, V> {
@@ -659,7 +659,7 @@ mod complex {
                 Test::eq(&v.cbrt(), &l.value)
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(any(feature = "std", feature = "libm"))]
             #[allow(trivial_casts)]
             fn mul_add(s: A<V>, a: A<V>, b: A<V>) -> bool {
                 #[allow(unused_imports)]
@@ -676,13 +676,13 @@ mod complex {
                 Test::eq(&s.mul_add(*a, *b), &r.value)
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(any(feature = "std", feature = "libm"))]
             #[allow(trivial_casts)]
             fn powi(v: A<V>) -> bool {
                 Test::eq(&v.powi(3), &Length::new::<meter>(*v).powi(P3::new()).value)
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(any(feature = "std", feature = "libm"))]
             #[allow(trivial_casts)]
             fn sqrt(v: A<V>) -> TestResult {
                 let l: Quantity<Q<P1, Z0, Z0>, U<V>, V> = Quantity::<Q<P2, Z0, Z0>, U<V>, V> {
```

wucke13:
@iliekturtles I applied the (unfortunately broken) patch, and rebased. Unfortunately some test still fail, now due to stack-overflows. It's not obvious to me what causes the stack-overflow. Possibly a recursion for the proxied functions, such as `crate::num::Float::atan` ?

iliekturtles:
Sounds like the proxy is calling itself instead of proxying the underlying function. Turbofish, like shown below, or explicitly using the full namespace to the function should resolve. I kicked off the builds to see which functions are having an issue.

https://github.com/iliekturtles/uom/blob/7c4b27b3cdc9530acd0c7116d5e073c1058207ce/src/lib.rs#L544-L546

wucke13:
But why do some tests succeeed without having the expicit `V as crate::num::Float`, while other pass? Where is the semantical difference between `asin` and `asinh`?

iliekturtles:
I tried running with `RUST_BACKTRACE=1` and most of the time it just hangs. Sometimes I get the stack overflow. Running without `RUST_BACKTRACE=1` always gives me the stack overflow. Possibly a `libm` or `rustc` bug? I didn't get much time to look at this issue today. Could you see if you could produce a minimal example?

iliekturtles:
I tried letting the run sit for a while with `RUST_BACKTRACE=1` and never got any output. I think the next step is attaching with a debugger and manually reviewing the stack.

wucke13:
I let an instance of `cargo test --verbose --no-default-features --features "autoconvert f32 si use_serde libm" -- --test-threads=1` run for 5 minutes now. It stuck on `si::angle::tests::trig::f32::atan2`. It hangs, but it does not overflow its stack and it has a constant memory utilization, resident memory is exactly `7316`. Still it consistently pulls 100% CPU load on a single core. What is this?

wucke13:
@iliekturtles I know understand the issue, I believe. Our problem is twofold. For one, quickcheck does weird things, when a test fails. But why do tests fail? We are checking the outputs of rusts std lib and `libm` for equivalence. But they are not equivalent. What do you think is the correct way forward here? Should we also use `libm` for the other side of the check if it is compiled in (instead of `std`)?

wucke13:
> I don't think that would explain the test hanging however.

Yes, I believe that is just https://github.com/BurntSushi/quickcheck/issues/295 I guess.f

> Yes, if libm is being used then we should make sure it's being used in tests as well. I don't think that would explain the test hanging however.

I'll try to implement it.

moritz-meier:
fix in https://github.com/rust-lang/libm/pull/275

iliekturtles:
Thanks for the latest update. Some errors in the build. I have a busy weekend coming up but will see if I can track down where they're coming from.

wucke13:
@iliekturtles Do you see any chance we get this merged in the closer future?

Issue #189, opened 2020-05-25
InformationRate * Time = Information != Ratio


exastion:
I'm pretty new to this library, but I think I found a bug concerning multiplication of different quantities.
Multiplying `InformationRate` and `Time` should result in `Information`, at least I would assume this and as far as I can see from the example, this type of multiplication should be possible.
Here a MVE that illustrates the problem:

```rust
use uom::si::f32::*;
use uom::si::information_rate::byte_per_second;
use uom::si::time::second;
use uom::si::information::byte;

fn main() {
    let r = InformationRate::new::<byte_per_second>(10.0);
    let t = Time::new::<second>(1.0);
    let b = r * t;
    let _ = b.get::<byte>();
}
```

According to the compiler error, it produces a `ratio`:
```sh
❯ cargo build    
   Compiling mve v0.1.0 (/home/sven/Documents/mve)
error[E0277]: the trait bound `uom::si::information::byte: uom::si::ratio::Unit` is not satisfied
  --> src/main.rs:11:15
   |
11 |     let _ = b.get::<byte>();
   |               ^^^ the trait `uom::si::ratio::Unit` is not implemented for `uom::si::information::byte`

error: aborting due to previous error
```

iliekturtles:
Thanks for the report. This is a known limitation and I believe specialization is needed to resolve. The current work around is to use `into()` to force the result's `Kind` associated type from the default that `Ratio` uses to `InformationKind`. e.g. `let b = (r * t).into();`

I'm leaving the issue open for the moment to review comments when `InformationRate` was implemented and see if there is any better solution that wouldn't require specialization.

PR #313, opened 2022-07-07
Add photometric units

quentinmit:
This PR adds `LuminousFlux` and `Illuminance` units. Unfortunately they are dimensionally equal to `LuminousIntensity` and `Luminance`. I tried to provide a `Mul` trait implementation to construct `LuminousFlux` from `LuminousIntensity * SolidAngle`, but try as I might I couldn't get the compiler to allow me to provide the trait. I saw #189 and related issues, but I don't think specialization is necessarily the solution - the compiler knows that `LuminousIntensity * SolidAngle` isn't allowed, so it shouldn't be a duplicate implementation... I dunno. Any advice is greatly appreciated.

### DIFF
--- a/src/si/illuminance.rs
+++ b/src/si/illuminance.rs
@@ -1,0 +1,61 @@
+//! Illuminance (base unit lux, lx, cd · sr / m²).
+
+quantity! {
+    /// Illuminance (base unit lux, lx, cd · sr / m²).
+    quantity: Illuminance; "illuminance";
+    /// Dimension of illuminance, E (base unit lux, lx, cd · sr / m²).
+    dimension: ISQ<
+        N2,     // length
+        Z0,     // mass
+        Z0,     // time
+        Z0,     // electric current
+        Z0,     // thermodynamic temperature
+        Z0,     // amount of substance
+        P1>;    // luminous intensity
+    kind: dyn (crate::si::marker::IlluminanceKind);
+    units {
+        @yottalux: prefix!(yotta); "Ylm", "yottalux", "yottaluxs";
+        @zettalux: prefix!(zetta); "Zlm", "zettalux", "zettaluxs";
+        @exalux: prefix!(exa); "Elm", "exalux", "exaluxs";
+        @petalux: prefix!(peta); "Plm", "petalux", "petaluxs";
+        @teralux: prefix!(tera); "Tlm", "teralux", "teraluxs";
+        @gigalux: prefix!(giga); "Glm", "gigalux", "gigaluxs";
+        @megalux: prefix!(mega); "Mlm", "megalux", "megaluxs";
+        @kilolux: prefix!(kilo); "klm", "kilolux", "kiloluxs";
+        @hectolux: prefix!(hecto); "hlm", "hectolux", "hectoluxs";
+        @decalux: prefix!(deca); "dalm", "decalux", "decaluxs";
+        /// The lux is defined to be 1 lumen per square meter,
+        /// or 1 candela steradian per square meter.
+        @lux: prefix!(none); "lm", "lux", "luxs";
+        @decilux: prefix!(deci); "dlm", "decilux", "deciluxs";
+        @centilux: prefix!(centi); "clm", "centilux", "centiluxs";
+        @millilux: prefix!(milli); "mlm", "millilux", "milliluxs";
+        @microlux: prefix!(micro); "µlm", "microlux", "microluxs";
+        @nanolux: prefix!(nano); "nlm", "nanolux", "nanoluxs";
+        @picolux: prefix!(pico); "plm", "picolux", "picoluxs";
+        @femtolux: prefix!(femto); "flm", "femtolux", "femtoluxs";
+        @attolux: prefix!(atto); "alm", "attolux", "attoluxs";
+        @zeptolux: prefix!(zepto); "zlm", "zeptolux", "zeptoluxs";
+        @yoctolux: prefix!(yocto); "ylm", "yoctolux", "yoctoluxs";
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    storage_types! {
+        use crate::si::quantities::*;
+        use crate::si::luminous_flux as lf;
+        use crate::si::area as a;
+        use crate::si::illuminance as i;
+        use crate::tests::{A, Test};
+
+        quickcheck! {
+            #[allow(trivial_casts)]
+            fn add(l: A<V>, r: A<V>) -> bool {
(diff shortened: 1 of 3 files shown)


iliekturtles:
Thanks for the PR! I'll review in detail over the next few days.

In regards to the `Mul` impl the problem is that the `Kind` associated type defaults to `uom::Kind`. I'm not sure/don't remember why the default isn't to maintain one of the operand's kinds -- if this even makes sense. I'll do some more digging into this as well but I'm not hopeful of a good short-term solution beyond `.into()`.

`type Output = ...` doesn't specify a kind which means the default is used.
https://github.com/iliekturtles/uom/blob/51a7cbf8238b0faa98b8f882614747d21cfb2f68/src/system.rs#L469-L484

quentinmit:
> In regards to the `Mul` impl the problem is that the `Kind` associated type defaults to `uom::Kind`. I'm not sure/don't remember why the default isn't to maintain one of the operand's kinds -- if this even makes sense. I'll do some more digging into this as well but I'm not hopeful of a good short-term solution beyond `.into()`.
> 
> `type Output = ...` doesn't specify a kind which means the default is used.

Naively, I would expect that by default the operators should only be provided for the case of `A<$kind> + B<$kind> = Output<$kind>`, so the example of `LuminousIntensity * SolidAngle` should be treated as `Kind * SolidAngleKind` and not have a `Mul` impl at all unless it's provided by a separate impl. By the way, the compiler error I get is:

```
error[E0308]: mismatched types
   --> src/si/luminous_flux.rs:109:26
    |
109 | /                          &(LuminousIntensity::<V>::new::<li::candela>((*l).clone())
110 | |                            * SolidAngle::<V>::new::<sa::steradian>((*r).clone())))
    | |_________________________________________________________________________________^ expected trait `LuminousFluxKind`, found trait `Kind`
```

so the compiler knows it's looking for a `LuminousFluxKind`. I was hoping that would make it select the alternate `Mul` implementation that can produce a `LuminousFluxKind`.

src/si/illuminance.rs
```diff
@@ -0,0 +1,61 @@
...
+        @exalux: prefix!(exa); "Elm", "exalux", "exaluxs";
+        @petalux: prefix!(peta); "Plm", "petalux", "petaluxs";
+        @teralux: prefix!(tera); "Tlm", "teralux", "teraluxs";
+        @gigalux: prefix!(giga); "Glm", "gigalux", "gigaluxs";
+        @megalux: prefix!(mega); "Mlm", "megalux", "megaluxs";
+        @kilolux: prefix!(kilo); "klm", "kilolux", "kiloluxs";
+        @hectolux: prefix!(hecto); "hlm", "hectolux", "hectoluxs";
+        @decalux: prefix!(deca); "dalm", "decalux", "decaluxs";
+        /// The lux is defined to be 1 lumen per square meter,
+        /// or 1 candela steradian per square meter.
```
iliekturtles, line 28:
```suggestion
        /// Dervied unit of illuminance. The lux is defined to be 1 lumen per square meter, or 1
        /// candela steradian per square meter.
```

src/si/illuminance.rs
```diff
@@ -0,0 +1,61 @@
...
+        @lux: prefix!(none); "lm", "lux", "luxs";
```
iliekturtles, line 29:
```suggestion
        @lux: prefix!(none); "lx", "lux", "lux";
```

Copy/paste error from luminous flux? It also seems the plural and singular are the [same](https://en.wikipedia.org/wiki/Lux).

src/si/luminous_flux.rs
```diff
@@ -0,0 +1,114 @@
...
+        @zettalumen: prefix!(zetta); "Zlm", "zettalumen", "zettalumens";
+        @exalumen: prefix!(exa); "Elm", "exalumen", "exalumens";
+        @petalumen: prefix!(peta); "Plm", "petalumen", "petalumens";
+        @teralumen: prefix!(tera); "Tlm", "teralumen", "teralumens";
+        @gigalumen: prefix!(giga); "Glm", "gigalumen", "gigalumens";
+        @megalumen: prefix!(mega); "Mlm", "megalumen", "megalumens";
+        @kilolumen: prefix!(kilo); "klm", "kilolumen", "kilolumens";
+        @hectolumen: prefix!(hecto); "hlm", "hectolumen", "hectolumens";
+        @decalumen: prefix!(deca); "dalm", "decalumen", "decalumens";
+        /// The lumen is defined to be 1 candela steradian.
```
iliekturtles, line 30:
```suggestion
        /// Derived unit of luminous flux. The lumen is defined to be 1 candela steradian.
```

iliekturtles (commented):
Finally got though a first pass at reviewing. I still haven't reviewed the `mul` impl in detail yet, but that is next on my list.

iliekturtles:
Only implementing `Mul` and `Div` for `Kind = uom::Kind` would require way too much work to implement a custom kind (10 `Mul` and `Div` implementations multiplied by all of the by-val by-ref combinations!) A simple implementation with generics `Kl` and `Kr` (LHS and RHS's kind) would require specialization.

What might be a possible path to pursue would be to implement `Mul` and `Div` for all `Kind * Kind` combinations. This would require the number of kinds squared impls (7<sup>2</sup> right now). With a macro this probably wouldn't be *too* painful:

```rust
impl_kind_ops!(Kind * Kind = Kind);
impl_kind_ops!(LuminousIntensityKind * Kind = Kind);
impl_kind_ops!(LuminousIntensityKind * SolidAngleKind = LuminousFluxKind);
// ...
```

What might be an issue is the `Mul`/`Div` impls for `Quantity`. I tried to do something similar for `Dimension` and I can't remember if things didn't work or if there was just an even larger explosion of type bounds that currently exists. The current mul implementation would be changed to use `$MulDivAlias` (a shortcut for `<A as Mul<V>>::Output` to determine what the output `Kind` should be instead of defaulting to `uom::Kind`.

```rust
impl<Dl, Kl, Dr, Kr, U, V> Mul<Quantity<Dr, U, V>> for Quantity<Dl, U, V>
where
    Dl: Dimension<Kind = Kl>,
    Dr: Dimension<Kind = Kr>,
    // ...
{
    type Output = Quantity<
        $quantities<$($crate::typenum::$AddSubAlias<Dl::$symbol, Dr::$symbol>,)+ Kind = $MulDivAlias<Kl, Kr>>,
         U, V>;

    fn mul(/*...*/) -> Self::Output {
        // ...
    }
}
```

Because of the scope of this work I'd like to see this effort separated from this PR. We can get `Illuminance` and `LuminousFlux` merged using the current functionality while finding out a better way to work with `Kind`s.

Possibly use #315 as the tracking issue and re-state the original problem.

PR #324, opened 2022-08-03
Add atomic units of energy, mass, length, electric charge and current.

crystal-growth:
Adding 

- Dalton (unified atomic mass unit = 1/12 * (mass of C12))
- Hartree (atomic unit of energy, ~27.2 eV)
- atomic unit of length (Bohr radius, 5.291 772 109 03 x 10-11 m )
- atomic unit of charge (absolute charge of electron, 1.602_176_634 x 10^-19 C)

PR #326, opened 2022-08-03
Add Electric Field quantity with corresponding units and test.

crystal-growth:
Adding Electric Field quantity [ElectricPotential / Length] with basic units V/m.

Additional non-SI units are:

-  Volts per mil
- atomic unit of electric field = Hartree Energy / electron charge / Bohr radius


crystal-growth:
Added test case for atomic unit of electric field, should fail until atomic units from PR #324 are merged.

PR #329, opened 2022-08-05
Add Electrical Mobility quantity and related units.

crystal-growth:
Adding electrical mobility:[Velocity / Electric Field, m2/v/s],   a ratio of charged particle average velocity in medium to the electric field.
Basic unit is m2/v/s, and a derived unit of cm2/v/s is regularly used in solid-state physics.


iliekturtles:
Thanks for the PR!

PR #331, opened 2022-08-05
Add ElectricalConductivity quantity with related units.

crystal-growth:
Adding electrical conductivity [ElectricalConductance / Length], with basic units siemens_per_meter and siemens_per_centimeter.


iliekturtles:
Accepted, thanks! If you don't mind deleting the old branches that have been merged that would help me a lot. I have your repo as a second remote in my local working copy and just seeing live branches would be a big help!

PR #333, opened 2022-08-05
Add ElectricCurrentDensity quantity with related units.

crystal-growth:
Adding electric current density {Current / Area} with base unit [A/m2]


iliekturtles:
Merged, thanks!

PR #334, opened 2022-08-06
Add NumberDensity quantities: Volumetric, Areal, and Linear.

crystal-growth:
Adding number density quantities:

- volumetric (1 per cubic meter )
- areal (1 per square meter)
- linear (1 per meter)

The kind of these quantities is concentration.


iliekturtles:
Added a bunch of units. Please review!

crystal-growth (commented):
Thank you for all this new units, I've rebased to master and squashed them.

PR #335, opened 2022-08-06
Add ReciprocalLength quantity with related units.

crystal-growth:
Adding ReciprocalLength quantity [1 / Length] with base unit 1/m.
Can be used for absorption coefficient, wavenumber, spatial frequency, etc


iliekturtles:
Thanks!

PR #343, opened 2022-08-10
Add MagneticMoment quantity with related units.

crystal-growth:
Adding magnetic moment quantity [Current x Area, or Energy / MagneticFluxDensity], A x m2, or Joule /Tesla.

Special units are Bohr magneton and Nuclear magneton, taken from CODATA2018.


iliekturtles:
Thanks!

PR #349, opened 2022-08-11
Add TemperatureGradient quantity with related units.

crystal-growth:
Adding temperature gradient quantity [TemperatureInterval / Length], with base unit Kelvin / meter.

I've checked that it returns TemperatureInterval when multiplied by length: 

`  let tgrad = TemperatureGradient::new::<kelvin_per_meter>(1.0);`
`  let l = Length::new::<meter>(2.0);`
`  let t_int: TemperatureInterval  = tgrad * l;` // ok

and fails (as it should) when trying to assign to ThermodynamicTemperature:

`    let t_abs: ThermodynamicTemperature  = tgrad * l;` // error[E0308]: mismatched types, ^^^^^^^^^ expected trait `TemperatureKind`, found trait `Kind`

Is just not providing "TempertureKind" a correct way of declaring quantities containing temperature interval in their dimension?

### DIFF
--- a/src/si/mod.rs
+++ b/src/si/mod.rs
@@ -113,6 +113,7 @@ system! {
         specific_heat_capacity::SpecificHeatCapacity,
         surface_electric_current_density::SurfaceElectricCurrentDensity,
         temperature_coefficient::TemperatureCoefficient,
+        temperature_gradient::TemperatureGradient,
         temperature_interval::TemperatureInterval,
         thermal_conductivity::ThermalConductivity,
         thermodynamic_temperature::ThermodynamicTemperature,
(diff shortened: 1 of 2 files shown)


iliekturtles:
In short, yes. However `Kind` is a pretty ugly wart in `uom` right now. See discussion in #313. Any multiplication or division loses kind information and defaults back to `uom::Kind`.

crystal-growth:
Thank you! 
Now I see why it is non-trivial to implement Entropy [J/K] and Molar Entropy [J/(K/*mol)] quantities, having the same dimension as heat capacity and molar heat capacity, but with absolute temperature instead of temperature interval.
Without specifying TemperatureKind they conflict with heat capacity and molar heat capacity.
With TemperatureKind -- they fail with mismatched type error since the kind disappears in all equatiuons like 
dS = dQ[Energy]/T[AbsoluteTemperature]

iliekturtles:
Exactly. The work-around right now is to use `result.into()` to force the result quantity into another quantity with the appropriate kind.

PR #355, opened 2022-08-16
Add volumetric, areal, and linear densities of states.

crystal-growth:
Adding density of quantum states for 3D, 2D, and 1D case, [NumberOfStates / (Volume * Energy)],
that is useful for solid state physics.
Basic units:
 1 / (m3*J) for volumetric DOS
 1 / (m2*J) for areal DOS
 1 / (m * J) for linear DOS


iliekturtles:
Thanks for the PR!

PR #358, opened 2022-08-21
Add "molar flow rate" units

crystal-growth:
Adding few "industrial" units of molar flow rate and corresponding units of AmountOfSubstance:

- standard centimeter per minute,
- standard liter per minute,
- standard cubic foot per minute,
- standard cubic meter per minute,

for standard conditions of 0C and 100000 Pa.

Sometimes some other "standard" conditions are assumed by hardware manufacturers, like T = 20C instead of 0C, or P = 1atm instead of 1 bar. 

The units were added to CatalyticActivity quantity, that now occupies dimension AmountOfSubstance/Time.
It may be convenient to add a type alias like MolarRate = CatalyticActivity.


iliekturtles:
Thanks!

PR #361, opened 2022-08-23
Add speed of light and Hartree atomic unit of velocity as velocity units

crystal-growth:
Adding CODATA values for speed of light (natural unit of velocity), and atomic unit of velocity.


iliekturtles:
Thanks for the PR!

PR #369, opened 2022-09-03
Add volumetric heat capacity quantity

crystal-growth:
Adding quantity for volumetric heat capacity [Energy / (TemperatureInterval * Volume)] 
with units

-  Joule per cubic meter kelvin
- Calorie per cubic meter kelvin


iliekturtles:
Thanks!

PR #377, opened 2022-09-23
Implement clamp

jacg:
`clamp` has two distinct implementations, one in `Ord` (which floats aren't) and another in `f{32,64}`.

I don't know how to separate these inside `uom`.


src/system.rs
```diff
@@ -1083,6 +1098,15 @@ macro_rules! system {
                     value: self.value.min(other.value),
                 }
             }
+
+            #[inline(always)]
+            fn clamp(self, min: Self, max: Self) -> Self {
```
iliekturtles, line 1103: This implementation isn't explicitly needed. `Ord` has a default implementation. I supose it's possible a manual implementation could be performant.

https://www.rustexplorer.com/b#%2F*%0A%5Bdependencies%5D%0Auom%20%3D%20%7B%20version%20%3D%20%22*%22%2C%20features%20%3D%20%5B%22i32%22%5D%20%7D%0A*%2F%0A%0Ause%20uom%3A%3Asi%3A%3Ai32%3A%3A*%3B%0Ause%20uom%3A%3Asi%3A%3Alength%3A%3A%7Bmeter%7D%3B%0A%0Afn%20main()%20%7B%0A%20%20%20%20let%20a%20%3D%20Length%3A%3Anew%3A%3A%3Cmeter%3E(15)%3B%0A%20%20%20%20let%20b%20%3D%20a.clamp(Length%3A%3Anew%3A%3A%3Cmeter%3E(17)%2C%20Length%3A%3Anew%3A%3A%3Cmeter%3E(20))%3B%0A%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20b)%3B%0A%7D

src/system.rs
```diff
@@ -754,6 +754,21 @@ macro_rules! system {
...
+            #[must_use = "method returns a new number and does not mutate the original value"]
+            #[inline(always)]
+            pub fn clamp(self, min: Self, max: Self) -> Self
+            where
+                V: $crate::num::Float,
+            {
+                Quantity {
+                    dimension: $crate::lib::marker::PhantomData,
+                    units: $crate::lib::marker::PhantomData,
+                    value: self.value.clamp(min.value, max.value),
```
iliekturtles, line 769: This is failing because it doesn't know where the `value.clamp` function is defined for the generic type `V`. The compiler is guessing the `Ord` trait because it is the only one with a `clamp` method. I see a couple solutions:

 * Use https://docs.rs/num/latest/num/fn.clamp.html. This may not be as efficient for `f32`/`f64`.
 * Create another `storage_types!` block just for float and then the code will compile

jacg: If you're going to raise the MSRV to 1.50.0, then I'd rather couple to the standard behaviour.

> Create another `storage_types!` block just for float 

You mean, alongside the one which already exists for `Float` and `Complex`?

If so, I'm getting something that compiles, but fails tests because of panics in the case that `max` < `min`. How do you put constraints on what quickcheck generates?

iliekturtles: Both my suggestions were in the context of raising the MSRV to 1.50. The first, using `num::clamp`, could be implemented in the `impl<D, U, V> Quantity<D, U, V> ...` block and would apply to all underlying storage types, however because it's using `num::clamp` it may not be as performant as a direct `f32::clamp` call. The second suggestion would be another `storage_types!` block that is just for `float` (the existing can't be used because `Complex` doesn't have a `clamp` method) and would essentially proxy `f32`/`f64` `clamp`. I lean towards the second option.

Reject invalid test input using `TestResult::discard();`:
https://github.com/iliekturtles/uom/blob/7c4b27b3cdc9530acd0c7116d5e073c1058207ce/src/tests/system.rs#L112-L114

jacg: Should I add some `#[should_panic]` tests? If so, I'm not sure how these fit in to the test strucre.

iliekturtles: Sure, something like `plant.tblResourcePackagingLinePackages` with `#[should_panic]` can be added as a non-quickcheck test up around line 263.

dnsco:
@iliekturtles Is there any way I can help get this merged?

The last failure seems to be about old MSRV

Issue #385, opened 2022-10-22
Support defmt framework


therealfrauholle:
We are using this crate heavily in an embedded context. The [defmt framework](https://crates.io/crates/defmt) is a well-known effort by ferrous systems to provide efficient logging and debugging for embedded devices. We wish to use uom together with this framework.

Arbitrary types can be logged through this framework, though they need to implement [`defmt::Format`](https://docs.rs/defmt/latest/defmt/trait.Format.html). So far we had to wrap uom values in a structure in order to log them, but I think it makes sense to provide this upstream. The new functionality should probably be feature gated similar to the `serde` functionality.

I will make a pull request for a proposal soon.

iliekturtles:
Sorry for the delays in catching up on these issues/PRs. PR for this feature is welcome. Let's discuss MSRV issues in #216.

iliekturtles:
MSRV bumped to 1.60.0 in #422.

asasine:
With MSRV at 1.65.0 as of 313bf68bcb5d45d3a7f9a5e03e66abfbb2648344, is this possible now?

CBJamo:
Would it be acceptable to have a separate MSRV for a defmt feature?

PR #386, opened 2022-10-22
Add defmt::Format implementation

therealfrauholle:
Solves #385 

This is a draft, since I still need to run the code. To my knowledge there is no easy way to test `defmt::Format` implementations.


therealfrauholle:
Cannot continue due to MSRV, will reopen when possible

PR #402, opened 2023-01-27
ARM floating point comparison tests

iliekturtles:


iliekturtles:
Currently broken into two commits with the second updating the `Test` impl for `Complex32` and `Complex64`. I'm not as confident with these changes because the original didn't fully pass.

PR #405, opened 2023-02-03
feat: Add ton_per_minute and ton_per_day

Uzaaft:
This PR attempts to add the ton_per_minute and ton_per_day units to flow rate.
Regarding the name of ton, I'd like to propose changing it to tonne, to avoid confusion with long and short ton. Or metric_ton.

PR #406, opened 2023-02-04
feat(volume_rate): Add cubic_meter_per_min and cubic_meter_per_hour as volume rate units

Uzaaft:
This PR attempts to add  cubic_meter_per_min and cubic_meter_per_hour as volume rate units.

### DIFF
--- a/src/si/volume_rate.rs
+++ b/src/si/volume_rate.rs
@@ -35,6 +35,10 @@ quantity! {
             "dam³/s", "cubic decameter per second", "cubic decameters per second";
         @cubic_meter_per_second: prefix!(none); "m³/s", "cubic meter per second",
             "cubic meters per second";
+        @cubic_meter_per_minute: 1.0_E0 / 6.0_E1; "m³/min", "cubic meter per minute",
+            "cubic meters per minute";
+        @cubic_meter_per_hour: 1.0_E0 / 3.6_E3; "m³/h", "cubic meter per hour",
+            "cubic meters per hour";
         @cubic_decimeter_per_second: prefix!(deci) * prefix!(deci) * prefix!(deci);
             "dm³/s", "cubic decimeter per second", "cubic decimeters per second";
         @cubic_centimeter_per_second: prefix!(centi) * prefix!(centi) * prefix!(centi);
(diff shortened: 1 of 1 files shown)


Uzaaft:
Waiting on #405  before marking this as ready for review.a

src/si/volume_rate.rs
```diff
@@ -35,6 +35,10 @@ quantity! {
             "dam³/s", "cubic decameter per second", "cubic decameters per second";
         @cubic_meter_per_second: prefix!(none); "m³/s", "cubic meter per second",
             "cubic meters per second";
+        @cubic_meter_per_minute: 1.0 / 60.0; "m³/min", "cubic meter per minute",
+            "cubic meters per minute";
+        @cubic_meter_per_hour: 1.0 / 3600.0; "m³/h", "cubic meter per hour",
+            "cubic meters per hour";
```
iliekturtles, line 41:
```suggestion
        @cubic_meter_per_minute: 1.0_E0 / 6.0_E1; "m³/min", "cubic meter per minute",
            "cubic meters per minute";
        @cubic_meter_per_hour: 1.0_E0 / 3.6_E3; "m³/h", "cubic meter per hour",
            "cubic meters per hour";
```

iliekturtles (commented):
Thanks for the PR! Changes look pretty good. Two minor changes below.

Uzaaft:
Thants for the suggestions, implemented those. :)
While I have you here @iliekturtles . I was wondering if you would be open to implement(or open for a contribution) that adds support for standard and normal conditions to the codebase?  
https://en.wikipedia.org/wiki/Standard_temperature_and_pressure

iliekturtles:
Similar to the other PR can you squash the commits and update the commit message. At the same time you can exclude the mass rate changes and just base this on top of iliekturles/master.

I'm definitely open to STP contributions. `AmountOfSubstance` already has a few STP specific units. Are you thinking about adding more or something different to more generally support STP?

Uzaaft:
> Similar to the other PR can you squash the commits and update the commit message. At the same time you can exclude the mass rate changes and just base this on top of iliekturles/master.
> 
> I'm definitely open to STP contributions. `AmountOfSubstance` already has a few STP specific units. Are you thinking about adding more or something different to more generally support STP?

well, what I ideally would want is to have a general support of STP wherever it could apply, as well as adding methods to convert from/into other units. I'll have to take a look at how `AmountOfSubstance` does it.

Uzaaft:
Everything should be ready now.

iliekturtles:
Thanks for the PR!

PR #415, opened 2023-05-20
Add support for micromath

Kaikallon:
Hello,

First off, I'm a beginner in Rust so I don't expect this PR to be accepted without modifications. Feedback is very welcome.

The reson for adding this support is that I wanted to use uom in an embedded environment without support for std. I need support for trigonometric functions, but I still want to keep the power of unit analysis. This is an attempt at doing that. 

One concern that I have about my contribution is that I think that there is too much copy pasting. I have played around with this for a while and I wasn't able to find a different way that works.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "uom"
-version = "0.34.0"
+version = "0.34.1"
 edition = "2018"
 authors = ["Mike Boutin <mike.boutin@gmail.com>"]
 description = "Units of measurement"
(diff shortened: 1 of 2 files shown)


Kaikallon:
I see now that num-traits is an optional feature of micromath. I'm not sure how or if that should be used. Probably because I never understood the purpose of num-traits. Many traits for numeric operations are already defined in the core and standard library.

Kaikallon:
Another thing that I noticed is that my implementation could be applied to ratio also. Further, there are more functions which could be implemented, such as sqrt()

iliekturtles:
Thanks for the PR! Another possible solution for trig functions on `no_std` is in #309 and uses `libm`. I'll compare the two (I just need some patience, I'm well behind on PRs) and see what kind of overlap there is.

I also kicked off the tests, but it looks like there is some bitrot and the latest compiler is causing some errors unrelated to your changes.

PR #431, opened 2023-08-08
Use a valid SPDX 2.1 license expression.

waywardmonkeys:


iliekturtles:
Thanks for the PR! Have any reference about "OR" being the proper way to license? I likely did a copy and paste in the past and am curious why "/" shouldn't be used.

waywardmonkeys:
Sure! Everyone used to do it that way with "/" as well.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields

iliekturtles:
Perfect, thanks for the reference. Change merged.

PR #432, opened 2023-08-08
msrv: Use rust-version key.

waywardmonkeys:
This was added in Rust 1.56 and is supported by the `cargo-msrv` tool as well.


iliekturtles:
Thanks for the PR. Merged!

PR #436, opened 2023-08-26
Add explicit serde feature

### LINKED ISSUE #216: Rename use_serde feature?
This would be a breaking change. See https://blog.turbo.fish/cargo-features/ for details how to enable this. I'm partial to the idea of a better feature name and `uom` is not yet committed to backwards compatibility, but this may end up being a larger hassle for users that it's worth.

Aehmlo:
This revision adds an explicit Cargo feature named `serde`. This feature is almost identical to the old `use_serde` feature, but uses features of Cargo's new feature resolver to automatically enable `serde` features for `num-*` crates as appropriate. The `use_serde` feature is now a (deprecated) alias for this new `serde` feature.

Very few changes were required, since `uom` was already using the (then-implicit) `serde` feature to begin with. I was unable to find an elegant way to warn users specifying `use_serde` to switch to the new feature.

This closes #216.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,7 +77,7 @@
 //!         "complex32", "complex64", # Complex floating point storage types.
 //!         "f32", "f64", # Floating point storage types.
 //!         "si", "std", # Built-in SI system and std library support.
-//!         "use_serde", # Serde support.
+//!         "serde", # Serde support.
 //!     ]
 //! }
 //! ```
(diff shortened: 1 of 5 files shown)


iliekturtles (commented):
Thanks for taking this on. One minor documentation change requested below.

src/lib.rs
```diff
@@ -77,7 +77,7 @@
 //!         "complex32", "complex64", # Complex floating point storage types.
 //!         "f32", "f64", # Floating point storage types.
 //!         "si", "std", # Built-in SI system and std library support.
-//!         "use_serde", # Serde support.
+//!         "serde", # Serde support.
```
iliekturtles, line 80: Could you add a note here and in README.md that the old `use_serde` feature is deprecated and will be removed no earlier than `v0.37.0`.

Aehmlo:
Added comments to Cargo.toml and README.md and squashed into the prior commit.

PR #437, opened 2023-08-26
Add defmt::Format support

### LINKED ISSUE #385: Support defmt framework

Aehmlo:
This is essentially #386 in a trenchcoat (thanks @therealfrauholle!), rebased against the latest upstream `master` and with a minor change to use an implicit `defmt` feature instead of an explicit one.

Closes #385.

### DIFF
--- a/.github/workflows/ci-full-test-suite.yml
+++ b/.github/workflows/ci-full-test-suite.yml
@@ -30,7 +30,10 @@ jobs:
           key: ${{ runner.os }}-${{ steps.install-rust.outputs.cachekey }}
 
       - name: Test all crates
-        run: cargo test --all --verbose --features "use_serde"
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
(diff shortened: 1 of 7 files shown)


Aehmlo:
It appears that I was confused and 1.60.0 isn't actually sufficient for this (currently looks like maybe 1.64.0 or even 1.67.0?). This may not be feasible with `defmt`'s current MSRV policy.

PR #453, opened 2024-01-15
Conversion now has both a value and a factor type

### LINKED ISSUE #452: Conversion to/from complex storage types is weong
Heya, i have run into the following issue: 

Lets assume a complex impedance like `Z = (123 + 321j) * ohm` 

```rust 
use uom::fmt::DisplayStyle;
use uom::si::complex64::ElectricalResistance;
use uom::si::electrical_resistance::ohm;
use uom::num_complex::Complex64;

fn main() {
    let n = Complex64::new(123.0,321.0);
    let unit = ElectricalResistance::new::<ohm>(n);
    println!("{} -> {}", n, unit.into_format_args(ohm,DisplayStyle::Abbreviation));
    assert_eq!(n,unit.get::<ohm>());
}
``` 

this fails: 
```
123+321i -> 343.7586362551492+0i Ω
thread 'main' panicked at src\main.rs:10:5:
assertion `left == right` failed
  left: Complex { re: 123.0, im: 321.0 }
 right: Complex { re: 343.7586362551492, im: 0.0 }
```

`343.7586362551492` being the norm of the number. 

I have looked at the implementing PR: while i would expect the pure unit conversion *factor* to be a floating point type, the actual conversion operation must be in the complex numbers, otherwise the operation is one-way (there are multiple solutions to z^2 = re^2 + im^2 for a given z).  

Im gonna have a go at either understanding where im wrong in my assumptions, or fixing this, as i really want complex impedance :D

mkalte666:
Initially, `ConversionFactor` and thus `Conversion::T` requred `PartialEq`. This makes sense for the conversion factor itself (i.e. scaling across units), however it breaks once you introduce complex numbers.

Those can *still* be scaled just like normal numbers - you essentially just increase or decrese a vector length, but the conversion function cannot compare them - "Z_1 < Z_2" is not trivially decidable.

It is, however, also not needed - unit scales are just that - scalars that scale. And those can be easily compared.

This commit seperates `Conversion::T` into `Conversion::VT` and `Conversion::T` and moves the `PartialEq` requirements from `ConversionFactor` into `Conversion::TT` directly.

This requires a lot of trait bounds added down the line, so im not 100% that this does not break anything down the line.

There might be a nicer way to go about this, but i haven't found any.

closes #452

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -408,13 +408,15 @@ pub enum ConstantOp {
 /// Trait to identify [units][units] which have a [conversion factor][factor].
 ///
 /// ## Generic Parameters
-/// * `V`: Underlying storage type trait is implemented for.
-///
+/// * `T`: The type of the conversion factor. Usually same as `VT`, but for example complex storage types have this as float as it needs `PartialEq`
+/// * `VT`: Underlying storage type trait is implemented for. Does not have to implement `PartialEq`, so its viable for complex data types.
 /// [units]: https://jcgm.bipm.org/vim/en/1.13.html
 /// [factor]: https://jcgm.bipm.org/vim/en/1.24.html
 pub trait Conversion<V> {
     /// Conversion factor type specific to the underlying storage type.
-    type T: ConversionFactor<V>;
+    type T: ConversionFactor<V> + PartialOrd;
+    /// Value type of the underlying type.
+    type VT: ConversionFactor<V> + From<Self::T>;
 
     /// Coefficient portion of [conversion factor](https://jcgm.bipm.org/vim/en/1.24.html) for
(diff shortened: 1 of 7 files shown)


iliekturtles:
Thanks for the PR. I kicked off the CI checks and will try to review this weekend.

mkalte666:
Should i rebase this and try again? Or do you have a different solution/idea/want for this? Not sure when i can schedule time again, but with any luck i can get it in before the weekend.

mkalte666:
Heya, i have rebased this to master and will put it into local testing again. Can you have a look at it at some point in the future @iliekturtles? Or if you need any input from my end (or me to come up with a better way of solving my complex number issue) i am also happy to help out, but this PR (excluding the rebase, working on reporting on it)  has been working for me so far

PR #457, opened 2024-01-27
Dev actions

iliekturtles:


.github/workflows/ci-full-test-suite.yml
```diff
@@ -10,15 +10,15 @@ jobs:
...
+        uses: actions/checkout@v4
 
       - name: Install Rust toolchain
         id: install-rust
         uses: dtolnay/rust-toolchain@stable
 
       # Cache based on OS/compiler.
       - name: Extract cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
```
adamreichold, line 21: You may want to consider [`rust-cache`](https://github.com/Swatinem/rust-cache) to replace the manual configuration here.

iliekturtles: Thanks for the suggestion. #458 opened.

PR #462, opened 2024-02-14
Fix abbreviations for kilogram per energy

einarhaaland:


iliekturtles:
Good catch. Thanks for the PR!

PR #463, opened 2024-02-22
Add `AngularMomentum`

### LINKED ISSUE #460: Built-In SI system has no alias for Angular Momentum
It's of course possible that I missed it, sorry if that's the case, but I can't seem to find a built-in type alias for angular momentum. It's easy enough to just define a type alias where needed, but it would be quite nice to have it included in the list of type aliases to begin with.

ISibboI:
Closes #460.

### DIFF
--- a/src/si/angular_momentum.rs
+++ b/src/si/angular_momentum.rs
@@ -1,0 +1,42 @@
+//! Angular momentum (base unit kilogram square meter per second, kg · m² · s⁻¹).
+
+quantity! {
+    /// Angular momentum (base unit kilogram square meter per second, kg · m² · s⁻¹).
+    quantity: AngularMomentum; "angular momentum";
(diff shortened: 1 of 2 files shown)


iliekturtles (commented):
Thanks for the PR! and sorry for my delay in responding. I've done a very brief review and things look really good. So far I only have a couple comments:

 * Are there other common units of angular momentum that can be added? Vary the mass and/or time portions?
 * Will you setup a `check_dimension` test and a `check_units` test for each individual unit. See [`acceleration`](https://github.com/iliekturtles/uom/blob/master/src/si/acceleration.rs) for an example.

ISibboI:
Thanks! I am sorry, but I don't have time to add functionality beyond what I need.

iliekturtles:
Thanks so much for the initial efforts. Closing this PR as #479 now supersedes this one.

PR #473, opened 2024-05-02
Fix incorrect `inverse_velocity` abbreviations

aevyrie:
This is a bug that I just noticed when upgrading from 0.33 to 0.36. This looks like a copy-paste error.


iliekturtles:
Thanks so much for the PR and sorry for the delay reviewing!

aevyrie:
No worries, thanks!

PR #477, opened 2024-07-03
docs: Replace relative URL by Rust item links

caspermeijn:
Some links in the documenation are broken. `cargo doc` will resolve item names when used in markdown links. The tool will report unresolved links as warning so that they can be fixed. Replace any relative URL by such item link. See https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

### DIFF
--- a/src/quantity.rs
+++ b/src/quantity.rs
@@ -121,7 +121,7 @@ macro_rules! quantity {
         pub type $quantity<U, V> = __system::Quantity<Dimension, U, V>;
 
         /// Marker trait to identify measurement units for the quantity. See
-        /// [`Unit`](../trait.Unit.html).
+        /// [`Unit`](__system::Unit).
         pub trait Unit: __system::Unit {}
 
         /// Trait to identify [units][units] which have a [conversion factor][factor] for the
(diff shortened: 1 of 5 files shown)


src/quantity.rs
```diff
@@ -354,12 +354,12 @@ macro_rules! quantity {
             /// Creates a struct that formats `self` for display.
             ///
             /// # Notes
-            /// Unlike [`format_args`](#method.format_args), the return value of this method can be
+            /// Unlike [`format_args`](Self::format_args()), the return value of this method can be
```
iliekturtles, line 357: These methods are included in the documentation for both `Quantity` (e.g. ) and individual `Quantity` type aliases (e.g. https://docs.rs/uom/latest/uom/si/absement/type.Absement.html#method.into_format_args). The new item links work for the `Quantity` page but on the type alias pages redirect to `Quantity` instead of the alias's documentation. Is there any way to keep these consistent without using the old style HTML links?

caspermeijn: Interesting! I didn't spot that.

It doesn't seem possible. Even `[`format_args`](Action::format_args())` resolves to `struct.Quantity.html#method.format_args`, which does make sense because Action is a type alias for Quantity.

Do you want me to change it back to using the original HTML link?

iliekturtles: Yeah, go ahead and change back.

caspermeijn: OK, I changed those back and checked that it links as expected.

iliekturtles:
Thanks for the PR! Finally merged!

PR #478, opened 2024-08-04
Fix NIST Special Publication 811 link

MKDJr:
I saw that the NIST Special Publication 811 link wasn't working and updated it to a working version: https://www.nist.gov/pml/special-publication-811/nist-guide-si-appendix-b-conversion-factors/nist-guide-si-appendix-b9.


iliekturtles:
Thanks so much!

PR #479, opened 2024-09-05
Add `AngularMomentum` (improved)

### LINKED ISSUE #460: Built-In SI system has no alias for Angular Momentum

Netzwerk2:
Closes #460 and is based on #463.

I implemented testing like you suggested and also added newton meter second as a unit (all the prefixes are copied from torque).
For now, I kept kilogram square meter per second, but I don't know if that's in your interest.

### DIFF
--- a/src/si/angular_momentum.rs
+++ b/src/si/angular_momentum.rs
@@ -1,0 +1,203 @@
+//! Angular momentum (base unit kilogram square meter per second, kg · m² · s⁻¹).
+
+quantity! {
+    /// Angular momentum (base unit kilogram square meter per second, kg · m² · s⁻¹).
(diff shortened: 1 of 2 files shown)


iliekturtles:
Thanks so much for the PR and sorry for the delay reviewing. I pushed formatting and some minor fixes to your fork. Could you please review and squash and then I'll look to get everything merged.

iliekturtles:
Merged! Thanks so much.

PR #483, opened 2024-09-30
Feat implemented support for schemars via a feature flag

akowi-sknobloch:
I've implemented the `JsonSchema` trait from the [schemars](https://crates.io/crates/schemars) crate.
I'm using serde + schemars in my project to generate a json schema so the users can easily create configs.
As i've switched a lot of my structs to use `uom` deriving the `JsonSchema` trait no longer worked.

I think supporting `schemars` would make a lot of sense for `uom`.
As `uom` already supports `serde` this addition would make building api's easier as you could simply generate your schema by deriving the trait.

My implementation of the Trait is pretty simple and covers the minimum.
I don't really like the schema name to be hard coded but i didn't know how i could set it automatically to eg. `Length`, `Mass`, etc.

Would be nice to get this feature merged :)
Happy for feedback and comments

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,25 +18,43 @@ autotests = true
 autobenches = true
 
 [package.metadata.docs.rs]
-features = ["usize", "u32", "u64", "isize", "i32", "i64", "bigint", "biguint", "rational", "rational32", "rational64", "bigrational", "serde"]
+features = [
+    "usize",
+    "u32",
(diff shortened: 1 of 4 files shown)


iliekturtles:
Thanks for the PR. I'll need to do some reading up on `schemars`. In the meantime I kicked off the automated tests to see if any immediate issues come up.

PR #512, opened 2025-04-07
Adding illuminance quantity.

### LINKED ISSUE #511: Include Illuminance and/or Irradiance as quantities?
'lo there.

I am trying to do some astro-calculations using uom. One quantity I would like to use is the, let's call it "brightness", of an object. There is a plethora of weird units for measuring brightness, which are a bit confusing to get acquainted with. The two that are most useful to me are [Illuminance](https://en.wikipedia.org/wiki/Illuminance) and [Irradiance](https://en.wikipedia.org/wiki/Irradiance). Both are currently not supported by the `uom::si` module, and either would suffice for my needs. The crate does offer the [Luminance](https://en.wikipedia.org/wiki/Luminance) type, but that is not well-defined for the point-like light sources that stars typically are.

Would you be willing to include one or two of them? Can and/or should I assist somehow?

TheComamba:
Adding illuminance quantity.
Closes #511.

### DIFF
--- a/src/si/illuminance.rs
+++ b/src/si/illuminance.rs
@@ -1,0 +1,120 @@
+//! Illuminance (base unit lux, cd · sr · m⁻²).
+
+quantity! {
+    /// Illuminance (base unit lux, cd · sr · m⁻²).
+    quantity: Illuminance; "illuminance";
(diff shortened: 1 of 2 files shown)


iliekturtles:
I pushed some minor changes (documentation formatting, code formatting, a few new units, additional tests) and have merged. Thanks for the PR!

PR #518, opened 2025-05-18
added alias for EnergyDensity

calbaker:


src/si/pressure.rs
```diff
@@ -96,6 +99,8 @@ quantity! {
             "pound-force per square inch", "pounds-force per square inch";
         @psi: 6.894_757_E3; "psi", "pound-force per square inch", "pounds-force per square inch";
         @torr: 1.333_224_E2; "Torr", "torr", "torr";
+        @joule_per_cubic_meter: prefix!(kilo) / prefix!(kilo); "J/m³",
```
calbaker, line 102: @iliekturtles , please scrutinize my use of `prefix!` here as I'm a little unsure about this.

jaskij:
Does `uom` put different quantities which use the same unit in the same module? I can't really find any examples right now, and it would seem confusing.

iliekturtles:
@calbaker, rather than defining a type alias, do you mind reworking the PR to have a new energy density quantity? See `SurfaceTension` and `RadiantExposure` as examples of quantities with the same base units.

PR #530, opened 2025-07-15
Add libm dependency

red-hara:
This commit adds the `libm` optional feature (not enabled by default) to turn the corresponding feature in the `num-traits` dependency crate. The commit updates some default SI units and changes `Float` exporting rules.

The `libm` crate enables math support for `#![no_std]` crates, including trigonometric functions. Those are implemented for `Angle` and `Ratio` quantities, and now are supported and usable in a wider area of cases.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "uom"
-version = "0.37.1"
+version = "0.37.2"
 edition = "2021"
 rust-version = "1.65.0"
 authors = ["Mike Boutin <mike.boutin@gmail.com>"]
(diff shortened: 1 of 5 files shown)


red-hara:
Thanks for pointing that out!

I see... I'm not the first one to stumble over lack of float-related operations in `uom`.

What would be the best strategy at the current moment? Should I invest some time into updating my changes in favor of suggested modifications in #309 or attempt to update the #309 to the current state of `uom`? (The possibility to rely on my own fork for now is an option too, but less preferable.)

That way or another, thanks for considering my pull request.

PR #538, opened 2025-08-17
Add clamp()

blaine-dodson:
UoM already has `min()` and `max()`, but seems to be missing `clamp()`. So I added it.


blaine-dodson:
closing in favor of #377

PR #543, opened 2025-08-30
Gyromagnetic ratio

SombkeMaximilian:
Adds support for the [gyromagnetic ratio](https://en.wikipedia.org/wiki/Gyromagnetic_ratio) (also known as magnetogyric ratio) quantity, i.e., the constant that relates a particle's magnetic moment to its angular momentum. This quantity is useful for describing, e.g., how magnetic dipoles precess in external magnetic fields and comes up for certain physics simulations and MRI/NMR data processing.

There are a few things I'm uncertain about in this PR. In the textbooks I've seen, the GR is often expressed in radians per second and tesla (from ħ = h / 2pi), but in the few practical applications that I've worked on, I only ever saw hz per tesla being used. I've added both, though I'm not completely sure about how your conversion coefficients work. I just wrote out the approximation of 2pi, which feels pretty dirty but I didn't find any example of pi being used in that context in the crate (very likely I didn't look closely enough).

I saw that some of the other quantities you've defined have a few useful constant multiples, so I added the most common ones that came up for me so far.

### DIFF
--- a/src/si/gyromagnetic_ratio.rs
+++ b/src/si/gyromagnetic_ratio.rs
@@ -1,0 +1,226 @@
+//! Gyromagnetic ratio, γ (base unit hertz per tesla, kg⁻¹ · s · A).
+
+quantity! {
+    /// Gyromagnetic ratio (base unit hertz per tesla, kg⁻¹ · s · A).
+    quantity: GyromagneticRatio; "gyromagnetic ratio";
+    /// Dimension of gyromagnetic ratio, M⁻¹TI (base unit hertz per tesla, kg⁻¹ · s · A).
+    dimension: ISQ<
+        Z0,     // length
+        N1,     // mass
+        P1,     // time
+        P1,     // electric current
+        Z0,     // thermodynamic temperature
+        Z0,     // amount of substance
+        Z0>;    // luminous intensity
+    units {
+        @yottahertz_per_tesla: prefix!(yotta); "YHz/T", "yottahertz per tesla",
+            "yottahertz per tesla";
+        @zettahertz_per_tesla: prefix!(zetta); "ZHz/T", "zettahertz per tesla",
+            "zettahertz per tesla";
+        @exahertz_per_tesla: prefix!(exa); "EHz/T", "exahertz per tesla", "exahertz per tesla";
+        @petahertz_per_tesla: prefix!(peta); "PHz/T", "petahertz per tesla", "petahertz per tesla";
+        @terahertz_per_tesla: prefix!(tera); "THz/T", "terahertz per tesla", "terahertz per tesla";
+        @gigahertz_per_tesla: prefix!(giga); "GHz/T", "gigahertz per tesla", "gigahertz per tesla";
+        @megahertz_per_tesla: prefix!(mega); "MHz/T", "megahertz per tesla", "megahertz per tesla";
+        @kilohertz_per_tesla: prefix!(kilo); "kHz/T", "kilohertz per tesla", "kilohertz per tesla";
+        @hectohertz_per_tesla: prefix!(hecto); "hHz/T", "hectohertz per tesla",
+            "hectohertz per tesla";
+        @decahertz_per_tesla: prefix!(deca); "daHz/T", "decahertz per tesla", "decahertz per tesla";
+        /// Derived unit of gyromagnetic ratio.
+        @hertz_per_tesla: prefix!(none); "Hz/T", "hertz per tesla", "hertz per tesla";
+        @decihertz_per_tesla: prefix!(deci); "dHz/T", "decihertz per tesla", "decihertz per tesla";
+        @centihertz_per_tesla: prefix!(centi); "cHz/T", "centihertz per tesla",
+            "centihertz per tesla";
+        @millihertz_per_tesla: prefix!(milli); "mHz/T", "millihertz per tesla",
+            "millihertz per tesla";
+        @microhertz_per_tesla: prefix!(micro); "µHz/T", "microhertz per tesla",
+            "microhertz per tesla";
+        @nanohertz_per_tesla: prefix!(nano); "nHz/T", "nanohertz per tesla", "nanohertz per tesla";
+        @picohertz_per_tesla: prefix!(pico); "pHz/T", "picohertz per tesla", "picohertz per tesla";
+        @femtohertz_per_tesla: prefix!(femto); "fHz/T", "femtohertz per tesla",
+            "femtohertz per tesla";
+        @attohertz_per_tesla: prefix!(atto); "aHz/T", "attohertz per tesla", "attohertz per tesla";
+        @zeptohertz_per_tesla: prefix!(zepto); "zHz/T", "zeptohertz per tesla",
+            "zeptohertz per tesla";
+        @yoctohertz_per_tesla: prefix!(yocto); "yHz/T", "yoctohertz per tesla",
+            "yoctohertz per tesla";
+
+        @radian_per_second_tesla: 1.0_E0; "rad/(s · T)", "radian per second tesla",
+            "radians per second tesla";
+
+        // Gyromagnetic ratio, γ
+        // γ = (magnetic dipole movement * nuclear magneton) / (spin * hbar)
+        // Magnetic dipole movement and spin https://nds.iaea.org/records/x2773-kj397
+        @proton_gyromagnetic_ratio: 2.675_221_870_8_E8; "γₚ", "proton gyromagnetic ratio",
+            "proton gyromagnetic ratio";
+        @neutron_gyromagnetic_ratio: 1.832_471_74_E8; "γₙ", "neutron gyromagnetic ratio",
+            "neutron gyromagnetic ratio";
+        @electron_gyromagnetic_ratio: 1.760_859_627_84_E11; "γₑ", "electron gyromagnetic ratio",
+            "electron gyromagnetic ratio";
+        @shielded_proton_gyromagnetic_ratio: 2.675_153_194_E8; "γ'ₚ",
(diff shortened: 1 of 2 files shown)


src/si/gyromagnetic_ratio.rs
```diff
@@ -0,0 +1,100 @@
...
+        @millihertz_per_tesla: prefix!(milli); "mHz/T", "millihertz per tesla", "millihertz per tesla";
+        @microhertz_per_tesla: prefix!(micro); "µHz/T", "microhertz per tesla", "microhertz per tesla";
+        @nanohertz_per_tesla: prefix!(nano); "nHz/T", "nanohertz per tesla", "nanohertz per tesla";
+        @picohertz_per_tesla: prefix!(pico); "pHz/T", "picohertz per tesla", "picohertz per tesla";
+        @femtohertz_per_tesla: prefix!(femto); "fHz/T", "femtohertz per tesla", "femtohertz per tesla";
+        @attohertz_per_tesla: prefix!(atto); "aHz/T", "attohertz per tesla", "attohertz per tesla";
+        @zeptohertz_per_tesla: prefix!(zepto); "zHz/T", "zeptohertz per tesla", "zeptohertz per tesla";
+        @yoctohertz_per_tesla: prefix!(yocto); "yHz/T", "yoctohertz per tesla", "yoctohertz per tesla";
+
+        @radian_per_second_tesla: 0.159_154_943_091_895_335_768_883_763_372_514_36; "rad/(s · T)", "radian per second tesla", "radians per second tesla";
```
iliekturtles: Shouldn't this conversion factor be `1.0` since rad/(s\*T) is equivalent to 1/(s\*T)?

The conversion factor used here appears to be a reduced gyromagnetic ratio (y_bar): `y = 2pi/y_bar`.

src/si/gyromagnetic_ratio.rs
```diff
@@ -0,0 +1,100 @@
...
+        @proton_gyromagnetic_ratio: 42.577_478_518 * prefix!(mega); "γ_n(proton)", "proton gyromagnetic ratio", "proton gyromagnetic ratios";
+        @deuton_gyromagnetic_ratio: 6.535_902_854 * prefix!(mega); "γ_n(deuton)", "deuton gyromagnetic ratio", "deuton gyromagnetic ratios";
+        @triton_gyromagnetic_ratio: 45.414_838_15 * prefix!(mega); "γ_n(triton)", "triton gyromagnetic ratio", "triton gyromagnetic ratios";
+        @boron11_gyromagnetic_ratio: 13.661_608_0 * prefix!(mega); "γ_n(boron11)", "boron11 gyromagnetic ratio", "boron11 gyromagnetic ratios";
+        @carbon13_gyromagnetic_ratio: 10.707_746 * prefix!(mega); "γ_n(carbon13)", "carbon13 gyromagnetic ratio", "carbon13 gyromagnetic ratios";
+        @nitrogen15_gyromagnetic_ratio: -4.315_255 * prefix!(mega); "γ_n(nitrogen15)", "nitrogen15 gyromagnetic ratio", "nitrogen15 gyromagnetic ratios";
+        @fluorine19_gyromagnetic_ratio: 40.069_244 * prefix!(mega); "γ_n(fluorine19)", "fluorine19 gyromagnetic ratio", "fluorine19 gyromagnetic ratios";
+        @phosphorus31_gyromagnetic_ratio: 17.241_162 * prefix!(mega); "γ_n(phosphorus31)", "phosphorus31 gyromagnetic ratio", "phosphorus31 gyromagnetic ratios";
+    }
```
iliekturtles, line 127: These appear to be "reduced gyromagnetic ratio" or ["gamma bar"](https://en.wikipedia.org/wiki/Nucleon_magnetic_moment#Nucleon_gyromagnetic_ratios) conversion factors instead of gyromagnetic ratios.

Why choose these specific nuclei? I see [wikipedia](https://en.wikipedia.org/wiki/Gyromagnetic_ratio#For_a_nucleus) has a larger table and [CODATA](https://physics.nist.gov/cgi-bin/cuu/Results?search_for=gyromagnetic+ratio) just includes electron, neutron, proton, and helion.

iliekturtles (commented):
Thanks for the PR! Sorry for the delay in reviewing.

src/si/gyromagnetic_ratio.rs
iliekturtles: EDIT: formula is actually `y = 2pi y_bar` / `y_bar = y/2pi`

src/si/gyromagnetic_ratio.rs
SombkeMaximilian: The context in which this unit came up was simulating nuclear magnetic resonance spectra. I included these nuclei because they are the commonly used one for organic samples (though it's mostly only 1H and 13C). Now that I do look at this again, though, calling them gyromagnetic ratios definitely seems weird, but it's what I found in the textbooks I looked at. In that wikipedia table, the middle column should be using the radians unit, while the right one is the one we use for NMR (divide by 2pi to get Hz per tesla).

Edit: After deliberating this a bit more again, this selection of nuclei is not particularly meaningful outside of NMR, so I wouldn't be opposed to not including these and avoiding the confusion with the reduced conversion factors.

src/si/gyromagnetic_ratio.rs
SombkeMaximilian: I have to admit that it has been a while since I looked at this, but I sort of just played around with the factors until I reached the correct values in calculations through trial and error. As mentioned in the initial comment of the PR, I just used an approximation of 1/(2 pi) for that because I couldn't find any other example of similar conversion factors.

src/si/gyromagnetic_ratio.rs
iliekturtles: I was doing a bit more reading and think it's valuable to have both the gyromagnetic ratio and the reduced gyromagnetic ratio values. The big question is if the reduced values should just be captured as separate units, or under a new quantity.

I have some local fixes that I still want to push so you can hold of making any new changes for the moment.

iliekturtles: I pushed some changes if you have a moment to review. I had reduced gyromagnetic ratios added as separate units currently. Reduced tests are failing because of significant figure issues.

SombkeMaximilian: I see nothing wrong. My other project (NMR simulations) still works with these changes. Some minor details I suppose:
- It might be worth it to include some other orders of magnitude for the angular frequency gyromagnetic ratio (radian per millisecond tesla, microsecond tesla, nanosecond tesla, etc.).
- Depending on which resources people use, the gyromagnetic ratio of specific nuclei or particles can differ. I usually use the IAEA or NNDC resources for these kinds of constants, but I've also seen others used in different fields. Just taking some of the common ones from these or CODATA is fine by me, though.

iliekturtles: Do you have references to conversion factors provided by IAEA or NNDC for these nuclei? I couldn't find them in limited searching. The wikipedia values are less precise and I was unable to actually access the references.

SombkeMaximilian: Here are the two we currently use:
https://doi.org/10.61092/iaea.yjpc-cns6
https://www.nndc.bnl.gov/nndc/stone_moments/nuclear-moments.pdf

Note that these experiments measure nuclear magnetic moments. The gyromagnetic ratio can then be obtained using the relation for the maximally aligned spin state:

magnetic moment = gyromagnetic ratio * spin * hbar

### Example

Particle: 13C
Spin: 0.5
Nuclear magnetic moment: 0.702_369 nm (nuclear magneton, from the IAEA source)

Angular gyromagnetic ratio (gamma): 6.72788e+7 rad / (s T)
Frequency gyromagnetic ratio (gamma / 2pi, what we use for NMR): 1.07077e+7 Hz / T

(Values calculated with WolframAlpha since it's been a while and I don't remember what I used to get the results in the original pull request)

Edit: The difference is probably related to the uncertainties, but I really can't remember what I did. If you want to keep conversions for specific nuclei, I would simply recalculate them using one of the tables above and constants (planck, nuclear magneton) from CODATA or similar.

iliekturtles: Thanks! I did find the IAEA magnetic dipole moments at one point, but was hoping to avoid the calculations. I'm hoping to review the values in your original PR and push my remaining changes by Friday... but we'll see how much time I find.

iliekturtles: Today isn't the Friday I was expecting!

I can't seem to get the same results from the magnetic moment formula. Solving for gyromagnetic ratio, the formula is `magnetic moment / spin / hbar`. But I don't get 6.72788_E7. Could you possibly provide more details?

SombkeMaximilian: I used wolfram alpha since I didn't remember how I calculated the values in my original pull request. Here's a link (you can also make it give you more digits, this was just as an example):

https://www.wolframalpha.com/input?i=N%5B%280.702369+nuclear+magneton%29+%2F+%280.5+hbar%29%2C+%7B1%2C+6%7D%5D

I don't know where they get their constants, but I would expect it to also just be CODATA. This (currently) yields 6.72788×10^7 ampere seconds per kilogram (or, equivalently, radians per second tesla) on my end.

iliekturtles: I finally took some time to review in more detail. The IAEA values are given in multiples of the [nuclear magneton](https://physics.nist.gov/cgi-bin/cuu/Value?mun|search_for=nuclear+magneton). I can finally reproduce and will fill in the remaining values!

iliekturtles:
Thanks for the PR and the patience. Merged!

PR #549, opened 2025-10-29
Add `data_mile` unit of length

jschillem:
Data miles are commonly used in fields involving radar technology. They are defined as a length of 6000 ft; converted to meters results in 1828.8 m.

For the abbreviation, there is no official abbreviation that I could find online or through any official documentation, though I have seen both "DM" and "data mi" used professionally while working with radars.


iliekturtles:
Thanks for the PR. Merged!

Issue #493, opened 2024-11-04
`Complex` unit tests are broken.


gechelberger:
ubuntu 24.04

I tried to bisect for the breaking commit, but I can't identify a "good" commit to work with.

Reproduction
```sh
# current master
git checkout d385d79
cargo clean && cargo +1.82.0 test -F complex32

# original complex release version
git checkout 51a7cbf
cargo clean && cargo +1.82.0 test -F complex32

# original release on msrv
cargo clean && cargo +1.65.0 test -F complex32

# original release on +nightly with min dependency versions
cargo clean && cargo minimal-versions test -F complex32
```

All of these build, but none of them pass the comlpex32 test-suite.

gechelberger:
[fail-d385d79-1.82.0-complex32.txt](https://github.com/user-attachments/files/17620914/fail-d385d79-1.82.0-complex32.txt)
[fail-51a7cbf-1.65.0-complex32.txt](https://github.com/user-attachments/files/17620915/fail-51a7cbf-1.65.0-complex32.txt)
[fail-51a7cbf-1.82.0-complex32.txt](https://github.com/user-attachments/files/17620916/fail-51a7cbf-1.82.0-complex32.txt)

PR #550, opened 2025-10-31
Fix conversion for complex units, add From, add thermal insulance, and some minor fixes

jokva:
Hi,

The main feature of these patches is fixing the conversion when working with units with complex values (real+imag component). Whenever bases would change uom would create values where the imaginary component was 0, and I'm not sure conversion ever worked.

---

Move the PartialOrd trait from the ConversionFactor and onto T. T is
usually f64 or something compatible (at the very least a scalar),
while V might be Complex. We really only need to compare the scaling
factor by the scalar (of type T). This didn't matter much when both V
and T where scalar, but it broke down for Complex.

This means Conversion<V> no longer has a default implementation for
conversion(), which must return something of the underlying type V.

It's quite common to start with some values that through
transformations end up complex, such as in signal processing. To
support this I added into_complex() (which is very clear in its
intention) and From<Quantity<V>> -> Quantity<Complex<V>>. This
supports code like:

    // m = meter, s = second
    let c1 = (m / s).into_complex();
    let c2 = uom::complex64::Velcity::from(m /s);
    assert!(c1 == c2);

into_complex() on an already.complex value is a no-op, to make it
easier to write generic code, e.g.:

    u.into_complex() == v.into_complex().into_complex()

The rest of the patch is dealing with the outcome; propagating trait
constraints and types. Notably, functions that call
{to,from,change}_base must provide +-*/ in order to support the
conversion.

There's still some work left, but complex types are at least usable
now, and more tests pass. What's missing:

1. The test suite needs some polish to better support
   complex, particularly the quickcheck tests. Scaling and
   transforming between bases when using very large values quickly
   overflow into inf or NaN, and the filters for this behaviour for
   non-complex numbers are not sufficient.
2. To/from string is not implemented. We probably don't want to use
   the default formatter for Complex, but rather output and parse
   a (a+bi) or (a+bj) format. This is not implemented and so these
   tests fail.
3. Some tests in si/ break because complex scaling does more floating
   point operations and introduce more imprecision. It might be
   sufficient to change some of these from eq to approx_eq.
4. The the rem/% operator which should probably be removed or at least
   carefully property checked for complex storage.

---

What would be nice is more general support for "promotion" between storage types. It would be neat if this code worked:

    let ms = (m / s) * (1.2 + Complex::i());
    // ms is now Velocity<Complex<float>>

I have not tried to support this use in this patch, but it's often the natural way to express formulae, rather than mucking about with From/Into or having to go via raw Unit::new() calls.

---

Finally, I fixed a few minor problems I encountered. I need a unit for thermal insulance/R-value, and some prefix'd units for areal heat capacity.

---

Please note that I made no real effort to fix all of the tests.

```
$ cargo --color=always test -F complex32 -F complex64

failures:
    si::areal_number_density::test::complex32::check_units
    si::areal_number_rate::test::complex32::check_units
    ... more check units
    tests::quantity::complex32::div_quantity
    tests::quantity::complex32::mul_quantity
    tests::quantity::complex32::rem
    tests::quantity::complex64::div_quantity
    tests::quantity::complex64::mul_quantity
    tests::system::complex32::to_base
    tests::system::complex::complex32::fp_categories
    tests::system::complex::complex64::fp_categories

test result: FAILED. 1209 passed; 29 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
```

This is much better than before, of course. The check units failures are of this kind:

```
thread 'si::volume_rate::tests::complex32::check_units' panicked at src/tests/mod.rs:165:28:
assertion `left == right` failed
  left: Complex { re: 0.0004719475, im: 0.0 }
 right: Complex { re: 0.00047194748, im: 0.0 }
```

The mul/div tests are of this kind:

```
thread 'tests::quantity::complex32::mul_quantity' panicked at quickcheck-1.0.3/src/tester.rs:165:28:
[quickcheck] TEST FAILED. Arguments: (A { v: Complex { re: -2.1490782e36, im: -4.325698e37 } }, A { v: Complex { re: 0.0, im: 7.2342386e37 } })
```
Most of them, and the spurious quickcheck failure, don't seem like proper failures to me -- rather, this is a result of broken relationships at the float boundaries and would probably go away if the float input space is more constrained.

### DIFF
--- a/src/si/mod.rs
+++ b/src/si/mod.rs
@@ -150,6 +150,7 @@ system! {
         temperature_interval::TemperatureInterval,
         thermal_conductance::ThermalConductance,
         thermal_conductivity::ThermalConductivity,
+        thermal_insulance::ThermalInsulance,
         thermal_resistance::ThermalResistance,
         thermodynamic_temperature::ThermodynamicTemperature,
         time::Time,
--- a/src/si/thermal_insulance.rs
+++ b/src/si/thermal_insulance.rs
@@ -1,0 +1,118 @@
+//! Thermal insulance (base unit square meter kelvin per watt, kg⁻¹ ⋅ m⁻² ⋅ s³ ⋅ K)
+
+quantity! {
+    /// Thermal insulance (base unit square meter kelvin per watt, kg⁻¹ ⋅ m⁻² ⋅ s³ ⋅ K).
+    /// Also called R-value.  Measures a materials resistance to the heat
+    /// current.
+    quantity: ThermalInsulance; "thermal insulance";
+    /// Dimension of thermal insulance, L⁻²M⁻¹T³Th (base unit square meter
+    /// kelvin per watt, kg⁻¹ ⋅ m⁻² ⋅ s³ ⋅ K).
+    dimension: ISQ<
+        Z0,     // length
+        N1,     // mass
+        P3,     // time
+        Z0,     // electric current
+        P1,     // thermodynamic temperature
+        Z0,     // amount of substance
+        Z0>;    // luminous intensity
+    units {
+        @square_meter_kelvin_per_yottawatt: prefix!(none) / prefix!(yotta); "m²K/YW",
+        "square meter kelvin per yottawatt", "square meter kelvins per yottawatt";
+        @square_meter_kelvin_per_zettawatt: prefix!(none) / prefix!(zetta); "m²K/ZW",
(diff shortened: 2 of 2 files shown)


jokva:
https://github.com/iliekturtles/uom/issues/493
https://github.com/iliekturtles/uom/issues/452

PR #553, opened 2025-11-10
Use `serde_core` instead of `serde`

Turbo87:
Replace the `serde` dependency with `serde_core` to improve build times. Since the crate only hand-implements Serialize/Deserialize traits without using derive macros, `serde_core` is sufficient and allows parallel compilation with `serde_derive`.

see https://github.com/serde-rs/serde/tree/v1.0.228/serde_core

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -299,7 +299,7 @@ dependencies = [
  "num-rational",
  "num-traits",
  "quickcheck",
- "serde",
+ "serde_core",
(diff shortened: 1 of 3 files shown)


iliekturtles:
Thanks for the PR. I rebased and simplified the `pub extern crate ...` + `pub use ...` into a single `pub extern crate ... as serde`. I'll plan to merge if CI checks all pass.

PR #568, opened 2026-09-10
Evaluate conversion factors at compile time

### LINKED ISSUE #60: Generate `Conversion<V>` implementations to return literals

okhsunrog:
Resolves #60.

For integer storage types, every `Quantity::new`/`get` approximated the `f64` factor literals with `Ratio::from_f64` (a continued fraction loop) at run time, multiplied the coefficients of all base units with `Ratio` arithmetic, and only then scaled the value. On targets without an FPU this pulls software `f64` and `num-rational` into every binary that converts units: a small USB-PD sink on a Cortex-M4F spends ~22 kB of its 65 kB `.text` on it (elagil/usbpd#52). This PR evaluates the complete conversion factor at compile time from the unit literals and scales the value with one integer multiply/divide. The run-time path stays as fallback, so results are unchanged. Measured on that example (thumbv7em, release, LTO): `.text` 65264 → 42116 bytes, no `__aeabi_d*` or `num_rational` symbols left.

- `Conversion` gets associated consts `COEFFICIENT` and `CONSTANT` holding the `f64` literals. `unit!` sets them; the default is `NAN` (unknown), so hand-written `Conversion` impls keep using the run-time path.
- New hidden `fraction` module with a `const fn` rational type. `Fraction::from_f64` mirrors `Ratio::from_f64` exactly (same continued fraction algorithm, same `T::max_value()` bound), so compile-time factors are identical to the run-time ones. A test compares them against `num_rational` on real coefficients.
- `ConversionFactor` gets `MAX_MAGNITUDE` and `scale()`. Integer storage types implement `scale` with the same cross-reduction as `Ratio::mul`, so overflow behavior is unchanged. Floats (which already fold well), `BigInt`, `BigUint`, `Ratio` and `Complex` keep the default `None` and the run-time path.
- `to_base`, `from_base` and `change_base` compute the factor in an inline `const` block and call `scale`, falling back to the existing code for units with a constant, factors that do not fit the storage type, or unknown literals.
- Tests for `Fraction` and for the compile-time factors of the test system.

Breaking changes

MSRV goes from 1.68.0 to 1.82.0 for floating point arithmetic in `const fn` and inline `const` blocks. `Conversion` and `ConversionFactor` gain defaulted items.

Testing

With integer storage features the test suite already fails on master (592 tests, e.g. `yotta*` factors that do not fit `u32`, CI only runs `--no-run` for them). This PR does not change that set: 591 failing with the patch, a strict subset, plus the new tests passing. The default, `f32 si`, and `bigint`/`complex` (`--no-run`) matrix passes, as do clippy, rustdoc, and `cargo check` on 1.82.0 with all features.


okhsunrog:
@iliekturtles can you take a look please?

